Issue #5, opened 2019-04-04
JS wrapper


apendleton:
We should create a Node wrapper using `neon`, with `neon-serde` for data interop. We should then write JS-side tests, possibly porting tests from existing carmen-cache

aarthykc:
After https://github.com/mapbox/carmen-core/pull/8 lands, next actions look like: 
- [ ] Write js wrappers for the writer and the reader functions
- [ ] Write tests that can be run by `npm test`
- [ ] Provide the ability to publish binaries

PR #8, opened 2019-04-10
Node bindings and Rust in the same repository

aarthykc:
Per #5 , we need to create a js wrapper for the reader and writer of the gridstore. 

## File organisation:
The three options to organise our files while keeping node and rust code in the same repository are: 

Option 1: Have all the rust work in the `/native` directory
Option 2: Have the rust code that deals with the gridstore under `rust-src` and the node binding code under `native` 
Option 3: Use `yarn workspace` or `lerna`  to separate the two - https://yarnpkg.com/en/docs/workspaces, https://medium.com/@jsilvax/a-workflow-guide-for-lerna-with-yarn-workspaces-60f97481149d. 

With Option 2, roughly our workflow would be to run `npm install` which runs neon build internally.  

--- 

## Update: 
We decided to go with Option 2 which means all the node-binding related files will live under the `native` directory and all the rust code that deals with the gridstore will live under the `rust-src`
directory. The reason why we didn't go with Option 1 or Option 3 is because: 
- Option 1: Didn't provide a differentiation between the rust code that deals with the gridstore specifically and the rust code associated with the node bindings for the library.
- Option 3: Yarn workspaces doesn't have a good way to use libraries as git dependencies in other projects without publishing a package using lerna. This is a very different workflow than the one that we're used to and I am less inclined to use this option while an easier, more familiar workflow is available for us to use (which is Option 2). 

### Next Actions: 
- [x] Pick an option - Option 2
- [x] Get the PR reviewed cc (@miccolis)
- [x] Mergetown

cc @miccolis, @apendleton, @Nmargolis


package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "node bindings for carmen-core",
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "carmen-core"
+  },
+  "author": "Aarthy Chandrasekhar <kcaarthy@gmail.com>",
+  "license": "MIT",
+  "dependencies": {
+    "neon-cli": "^0.2.0"
```
miccolis: Should this be a devDependency?

aarthykc: 👌, moved to devDependencies!

aarthykc:
Added a few more next actions to - #5

PR #12, opened 2019-05-01
Better matching text

apendleton:
This PR largely rewrites `get_matching` as it exists in master, to actually return results in a usefully sorted order, rather than in whatever order it happens to encounter them during its database scan. The main implementation here is not geospatially aware; it returns things ordered descending by `(matches_language, relev, score, (x-y z-order composite), id)`.

The approach boils down to:
* scan the database for all entries within the range, load all flatbuffers structures, and bucket them into the ones that match the requested language and the ones that don't
* for each language-match bucket, eagerly scan all flatbuffers entries for all relev/score pairs they each contain, and bucket them across flatbuffers together across flatbuffers entries (so the output will be a map that goes from (relev/score) => (all coord lists across flatbuffers entries that share that relev/score)) -- this is eager for simplicity's sake, since there can be at most 32 buckets, and 32 entries per flatbuffers entry (4 relev values * 8 score values)
* for each of *those* bins, from each entry there can be lots and lots of x/y pairs (as many as there can be tiles at any given zoom level) so we merge the different lists of x/y pairs for any given relev/score across entries lazily, using the k-way-merge implementation from the `itertools` crate
* for each of *those*, we need to merge together all the IDs across phrase entries for a given relev/score/x/y combination. Here again, it seems like the number of possible IDs that could be present for a given set of phrases and relevance and score in a given tile is bounded, so we can be eager again, and just pile them into a vector and sort it

Also:
* I added a notion of `MatchOpts` that makes it possible to communicate bbox and proximity into `get_matching`, and wrote a really simple/stupid bbox implementation to show how this could work -- it doesn't do any of the binary search fanciness we actually want to do, and instead just runs the global one and then filters it for the given box. We should replace this with a real implementation once we decide what it should do. The MatchOpts also has a slot for proximity, but it's currently ignored, so the `match` in `get_matching` that looks at MatchOpts should eventually have four possibilities (proximity, bbox, neither, both) instead of two.

Plus I added a bunch of tests.


rust-src/src/gridstore/mod.rs
```diff
@@ -105,3 +105,151 @@ fn score_test() {
...
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 1, x: 24, y: 1, id: 12, source_phrase_hash: 0 }, matches_language: false }
+        ]
+    );
+
+    let search_key = MatchKey { match_phrase: MatchPhrase::Range { start: 1, end: 3 }, lang_set: 0 };
+    let records: Vec<_> = reader.get_matching(&search_key, &MatchOpts::default()).unwrap().collect();
+    assert_eq!(
+        records,
+        [
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 7, x: 58, y: 1, id: 30, source_phrase_hash: 0 }, matches_language: false },
```
aarthykc: @apendleton curious what the sort order is supposed to be here?

PR #13, opened 2019-05-02
CI and coverage

apendleton:
This PR lays out the building blocks for running CI and coverage on this repo. What I've got:

- [x] a Travis build matrix that runs both Rust and Node tests
- [x] a node "test" that just loads the module and ensures that it doesn't die; we can swap this out in the future for actual tests
- [x] on the Rust side, we run tests and also capture coverage
- [x] Rust coverage in this PR is with kcov/cargo-kcov -- I'm not wild about this, as its results seem kind of crappy (and inclined towards false negatives, in particular). My preferred coverage tool is [tarpaulin](https://github.com/xd009642/tarpaulin) (which we've used for fuzzy-phrase), but in addition to generally being a pain to get working, it seems like we're hitting an issue with the current release that causes our tests to intermittently segfault -- see issue 190 in that repo. I might revisit in the future once that's fixed; in the meantime, I'll probably enable coverage but not enforce it since the results from kcov are mediocre.
- [x] Rust formatting using `rustfmt`

That means the build matrix is two Node testers (one Mac, one Linux, both Node 10), and one Rust tester (stable only) which does regular tests, coverage, and rustfmt.

To-dos either on this PR, or more likely, later ones:
* set up pre-publishing of binaries on the Node side, using `node-pre-gyp`
* add JS-side linting, once we actually have some JS code to lint


apendleton:
Per chat, @Nmargolis will make the npm --> yarn update in #15 so going to merge this as-is.

PR #14, opened 2019-05-03
Use read-only open for rocksdb in the grid reader

apendleton:
We want to use read-only opening for rocksdb in our grid store at query time, so we don't get lock contention when multiple geocoding processes try to access the same grid data at the same time. This is not supported by rust-rocksdb master, though it *is* supported by the lower-level unsafe direct wrapper it includes, librocksdb-sys. There's a PR under development on the rust-rocksdb side that adds support, but does it as part of a major refactor/API change that looks like it might take a bit.

I looked at if I could use the low-level bindings and then do some unsafe shenanigans to turn a read-only low-level database handle into a high-level rust-rocksdb handle -- that was my original plan. Turns out I think this would be possible, but would involve lots more code copying that I initially thought. It ended up being much easier to fork rust-rocksdb and add support there, so that's what I did: https://github.com/apendleton/rust-rocksdb/tree/basic-ro . This adds support as minimally as possible (it's a ~15-line change) and we should be able to track master until the real support lands. It does not match the API that will eventually be provided, and also does not provide the seatbelts the real version ultimately will -- from what I can tell, read-only databases in the thing that will ultimately get released just won't have write methods, whereas in my branch they're still there, and using them results in a process crash. I think we can live with that limitation for now, though -- it's the same as in current carmen-cache.


apendleton:
@miccolis I added comments explaining the use of this fork (and other forks as well). Should be good to re-review.

Issue #1, opened 2019-04-04
Build out the grid store in Rust


apendleton:
We want to build a new storage backend for carmen, replacing the existing grid cache and fixing a bunch of its deficiencies. That work has started; this will be the tracking ticket for progress on it.

Differences from the existing one:
* stores data in a nested structure with like values grouped together, rather than a flat structure that's then delta-encoded
* doesn't pack things into 53-bit integers
* uses more bits to represent some fields for higher fidelity
* uses phrase IDs as the keys rather than phrase strings
* represents coordinates using [Z-order curve](https://en.wikipedia.org/wiki/Z-order_curve) representation, so that geographic seek is possible and spatially proximate entries are near one another in the data representation
* Use [FlatBuffers](https://en.wikipedia.org/wiki/FlatBuffers) instead of protocol buffers so that random seek within a buffer is possible without deserializing the whole thing or walking it in order
* Use the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) we've established in Fuzzy Phrase: there's one class for accumulating data to be written during index creation that's write-only, and a different class for reading and interacting with data at query time that's read-only. The current implementation has two classes that can both be used for reading, meaning all the read functions have to be implemented twice, and conformance tests are necessary to make sure they behave the same way (though in reality they still have subtle differences in behavior). This is a waste of effort.

### Initial builder work
- [x] create internal temp storage structure
- [x] create set/insert functions that write to temp storage
- [x] write the flatbuffers format spec for writing data to the database
- [x] figure out rocksdb integration
- [x] write a finish/pack function that can write the data from the temp storage into the rocksdb database in the specified flatbuffers format

### Reader
- [x] write logic for loading a builder
- [x] write get function
- [x] write list function
- [x] write basic, non-prox/non-bbox getmatching function
- [x] write prox-aware getmatching strategy that takes advantage of z-order
- [x] write bbox-aware getmatching strategy that takes advantage of z-order
- [x] (maybe) write getmatching strategy for when both are in play? or can we just use one or the other?

### More stuff
- [x] figure out how to emulate the current cached-prefix logic
[...]

PR #17, opened 2019-05-08
Spatial binary search over grids

aarthykc:
## Context 
When the user provides a bounding box or a proximity parameter, we need a way to search over the grids and return those that fall within the bbox and near the proximity point. This PR adds a binary search over the grids and integrates it with the worked that moved as a part of #1 

## Summary of changes 
- Adds `rust-src/src/gridstore/spatial.rs` - core logic for binary search over grids + tests 
- Modifies `rust-src/src/gridstore/store.rs` - integrates bbox_filter with current logic

## Next actions: 
- [x] Wire up bbox_filter with `store.rs`
- [x] Write remaining tests outlined here - https://github.com/mapbox/carmen-core/compare/spatialsearch?expand=1#diff-323bb3f1330715ac0db81b04e910cf1bR38
- [ ] Review
- [ ] Merge 

## Maybe now, maybe follow up

- [ ] Avoid deinterleaving morton order twice
- [ ] Optimize binary search for min/max use case.
- [ ] Hoist bbox interator up out of the generic get_matching

cc @apendleton @Nmargolis @miccolis

### DIFF
--- a/rust-src/src/gridstore/spatial.rs
+++ b/rust-src/src/gridstore/spatial.rs
@@ -3,29 +3,41 @@ use crate::gridstore::gridstore_generated::*;
 use morton::interleave_morton;
 use std::cmp::Ordering::{Less, Equal, Greater};
 
-fn bbox_filter<'a>(coords: &'a flatbuffers::Vector<'a, flatbuffers::ForwardsUOffset<Coord>>, bbox: [u16; 4]) -> impl Iterator<Item=Coord<'a>> {
+pub fn bbox_filter<'a>(coords: flatbuffers::Vector<'a, flatbuffers::ForwardsUOffset<Coord>>, bbox: [u16; 4]) -> impl Iterator<Item=Coord<'a>> {
     let min = interleave_morton(bbox[0], bbox[1]);
     let max = interleave_morton(bbox[2], bbox[3]);
-    let start = bbox_binary_search(&coords, min, 0).unwrap();
-    let end = bbox_binary_search(&coords, max, start).unwrap();
+    let start = match bbox_binary_search(&coords, min, 0) {
+        Ok(v) => v,
+        Err(v) => v,
+    };
+    let end = match bbox_binary_search(&coords, max, start) {
+        Ok(v) => v,
+        Err(v) => v,
+    };
+    assert!(start <= end, "start is before end");
     (start..end).map(move |idx| coords.get(idx as usize))
 }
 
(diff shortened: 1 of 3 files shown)


rust-src/src/gridstore/spatial.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    let range_start = coords.get(0).coord();
+    if min > range_start {
+        return None;
+    }
+    let range_end = coords.get(len - 1).coord();
+    if max < range_end {
+        return None;
+    }
+    debug_assert!(range_start.cmp(&range_end) != Less, "Expected descending sort");
```
apendleton: Any reason this can't be spelled `range_start > range_end`?

apendleton: Actually, looks like this pattern looks like it recurs, so I assume there's a reason for all of them, but if not, the `.cmp()` seems harder to parse than regular comparison operators

rust-src/src/gridstore/store.rs
```diff
@@ -259,13 +243,27 @@ impl GridStore {
                     // mask for the least significant four bits
                     let score = relev_score & 15;
 
-                    let coords =
+                    let coords_vec =
                         get_vector::<Coord>(record_ref.1, &rs_obj._tab, RelevScore::VT_COORDS)
                             .unwrap();
+                    let coords = match match_opts {
+                        MatchOpts { bbox: None, .. } => {
```
apendleton: So... I get the benefit from a code-sharing perspective of just having one get-matching function and moving the check here, rather than having a separate function for the bbox version, but what it ultimately means is, we now branch on whether or not the bounding box is set once for each relev/score for each phrase, rather than once, ever. If there's a way to hoist this check out of two levels of loop nesting, we should do it, especially as it would also let us avoid the indirection necessary from having to box the iterator. Can wait for a future PR if that's preferable.

rust-src/src/gridstore/spatial.rs
miccolis: I think I started doing this b/c the binary_search_by implementation does it this way. But I agree it's confusing. I'll switch anything that wasn't pulled directly over from there.

rust-src/src/gridstore/store.rs
miccolis: I agree on all points. So, let's get this in and then I'll take a pass at hoisting creation of closure that can kick off the appropriate iterator based on bbox (or not)

PR #19, opened 2019-05-14
Implement proximity filtering and proximity bbox filtering

aarthykc:
When the user provides a proximity parameter, we need a way to search over the grids and return those that fall near the proximity point. When a user provides both - a proximity point and a bbox filter we should return points within the bbox and near the proximity point provided. This PR adds the ability to sort the grids based on the proximity point and also provides the option to filter by bbox and sort by proximity point by providing both parameters. 

## Changes: 
- Adds `proximity()` function to `spatial.rs`
- Adds `bbox_proximity_filter()` function to `spatial.rs`
- Adds unit tests for both 

## Next actions: 
- [x] Write more unit tests 
- [x] Review 
- [x] Integrate with `store.rs` 

cc @apendleton, @miccolis, @Nmargolis

### DIFF
--- a/rust-src/src/gridstore/mod.rs
+++ b/rust-src/src/gridstore/mod.rs
@@ -130,6 +130,7 @@ fn matching_test() {
                 GridEntry { id: i + 3, x: ((2 * i) + 1) as u16, y: 1, relev: 1., score: 7, source_phrase_hash: 0 },
             ];
             i += 4;
+
             builder.insert(&key, &entries).expect("Unable to insert record");
         }
     }
--- a/rust-src/src/gridstore/mod.rs
+++ b/rust-src/src/gridstore/mod.rs
@@ -130,7 +130,6 @@ fn matching_test() {
                 GridEntry { id: i + 3, x: ((2 * i) + 1) as u16, y: 1, relev: 1., score: 7, source_phrase_hash: 0 },
             ];
             i += 4;
-
             builder.insert(&key, &entries).expect("Unable to insert record");
         }
     }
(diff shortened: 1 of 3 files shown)


rust-src/src/gridstore/spatial.rs
```diff
@@ -55,6 +68,60 @@ pub fn bbox_filter<'a>(
     }))
 }
 
+pub fn proximity_filter<'a>(
```
miccolis: Naming nitpick, this isn't filtering - just iterating in a different order - we should probably pick a different name.

aarthykc: changed `proximity_filter` -> `proximity`

rust-src/src/gridstore/spatial.rs
```diff
@@ -55,6 +72,68 @@ pub fn bbox_filter<'a>(
     }))
 }
 
+/// Generate an Iterator over a Coord Vector given a proximity point
+///
+/// Returns [`Some(Iterator<>`] which is a Coord Vector morton order range that overlaps with a bounding box and is ordered by the z-order distance from the proximity point
+/// [`None`] if the bounding box does not overlap with the morton order range
```
miccolis: This comment talks about a bounding box, but this function doesn't have one.

rust-src/src/gridstore/spatial.rs
```diff
@@ -44,8 +44,25 @@ pub fn bbox_filter<'a>(
...
+    coords: flatbuffers::Vector<'a, flatbuffers::ForwardsUOffset<Coord<'a>>>,
+    bbox: [u16; 4],
+) -> Option<impl Iterator<Item = Coord<'a>>> {
+    let len = coords.len();
+    if len == 0 {
+        return None;
+    }
+
+    let range = bbox_range(coords, bbox)?;
+    Some((range.0..(range.1 + 1)).filter_map(move |idx| {
```
miccolis: Should we use the inclusive range here? Like `(range.0..=range.1)`?

rust-src/src/gridstore/spatial.rs
aarthykc: @miccolis good catch, I flipped proximity and bbox_proximity

rust-src/src/gridstore/mod.rs
```diff
@@ -130,7 +130,6 @@ fn matching_test() {
                 GridEntry { id: i + 3, x: ((2 * i) + 1) as u16, y: 1, relev: 1., score: 7, source_phrase_hash: 0 },
             ];
             i += 4;
-
```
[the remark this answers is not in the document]
aarthykc: Done!

rust-src/src/gridstore/spatial.rs
aarthykc: Sure!

rust-src/src/gridstore/mod.rs
```diff
@@ -297,4 +296,53 @@ fn matching_test() {
...
+    assert_eq!(
+        records,
+        [
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 7, x: 26, y: 1, id: 14, source_phrase_hash: 0 }, matches_language: true },
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 7, x: 25, y: 1, id: 15, source_phrase_hash: 0 }, matches_language: true },
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 7, x: 25, y: 1, id: 13, source_phrase_hash: 0 }, matches_language: true },
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 1, x: 24, y: 1, id: 12, source_phrase_hash: 0 }, matches_language: true },
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 7, x: 41, y: 1, id: 23, source_phrase_hash: 0 }, matches_language: false },
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 7, x: 41, y: 1, id: 21, source_phrase_hash: 0 }, matches_language: false },
+            MatchEntry { grid_entry: GridEntry { relev: 1.0, score: 7, x: 42, y: 1, id: 22, source_phrase_hash: 0 }, matches_language: false },
```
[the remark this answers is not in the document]
aarthykc: @miccolis, that is totally a bug, thanks for checking!

rust-src/src/gridstore/spatial.rs
miccolis: @aarthykc I don't see this change...

rust-src/src/gridstore/spatial.rs
```diff
@@ -55,6 +72,78 @@ pub fn bbox_filter<'a>(
...
+    if coords.len() == 0 {
+        return None;
+    }
+
+    let prox_mid = match coord_binary_search(&coords, prox_pt as u32, 0) {
+        Ok(v) => v,
+        Err(_) => return None,
+    };
+
+    let getter = move |idx| {
```
miccolis: Can we change this to `filtered_get` or similar?

PR #20, opened 2019-05-15
Coalesce

Nmargolis:
### Context
This PR is essentially a port of [coalecsce from carmen-cache](https://github.com/mapbox/carmen-cache/blob/master/src/coalesce.cpp), with some reorganization, and a rewrite of the scaling proximity and bbox to different zoom levels.

This first part of the work is ready to merge so node bindings can be built on top of it. There is more work to do to optimize performance, tidy up and add [test coverage](https://github.com/mapbox/carmen-core/pull/21).

### Summary of changes 
- [x] Add coalesce function
- [x] Rewrite the `pxy2zxy` and `bxy2zxy` functions from carmen-cache since they weren't working properly in carmen-cache, as `MatchOpts.adjust_to_zoom()`
- [x] Reformat the MatchOpts to enforce a matching zoom on the bbox and proximity options

### Next steps
- [ ] Add tests for coalesce, moving in #21 
- [ ] Clean up todos
- [ ] Add optimization to use previous grid's relev, distance and scoredist if inputs are the same, instead of re-calculating
- [ ] Only call adjust_to_zoom if zooms are different


rust-src/src/gridstore/coalesce.rs
```diff
@@ -0,0 +1,442 @@
...
+    contexts.dedup_by_key(|context| context.entries[0].grid_entry.id);
+    contexts.truncate(MAX_CONTEXTS);
+    Ok(contexts)
+}
+
+fn coalesce_multi(
+    stack: &[PhrasematchSubquery],
+    match_opts: &MatchOpts,
+) -> Result<Vec<CoalesceContext>, Box<Error>> {
+    let mut stack: Vec<PhrasematchSubquery> = stack.iter().cloned().collect();
```
apendleton: So... I definitely see why this is necessary, but it's kind of a bummer. I think the pattern that will almost definitely happen in practice will be that in the JS wrapper we'll copy everything from JS land into a Rust vector, then send a reference to that vector into coalesce, then copy everything into another vector here so we can sort it, and that extra copy doesn't accomplish anything, especially because we aren't going to reuse the vector of subqueries for anything else, so there's probably not any harm in just doing a move.

So: despite the fact that it was me who suggested the `stack: &[PhrasematchSubquery]` signature, I'm now thinking it should probably instead be `mut stack: Vec<PhrasematchSubquery>` for both this function and regular `coalesce`, so that we can just sort it in-place and avoid the extra copy. (`coalesceSingle` could either stay as is or change to `subquery: PhrasematchSubquery`, I don't think it really matters.

rust-src/src/gridstore/coalesce.rs
```diff
@@ -0,0 +1,442 @@
...
+                    coalesced.insert(
+                        zxy,
+                        vec![CoalesceContext { entries, mask: context_mask, relev: context_relev }],
+                    );
+                }
+            }
+        }
+    }
+
+    for (_, matched) in coalesced {
```
apendleton: This could also be `for matched in coalesced.values()`

rust-src/src/gridstore/coalesce.rs
```diff
@@ -0,0 +1,442 @@
...
+
+fn scoredist(mut zoom: u16, mut distance: f64, mut score: u8, radius: f64) -> f64 {
+    if zoom < 6 {
+        zoom = 6;
+    }
+    if score > 7 {
+        score = 7;
+    }
+
+    // If the distance is 0, set a minimum distance to avoid dividing by distatios that approach zero
```
apendleton: comment typo? `distatios`

rust-src/src/gridstore/coalesce.rs
Nmargolis: Makes sense. It seems that it's not necessary to make stack mutable in the coalesce signature. Maybe since it's transferring ownership to coalsece_multi, and coalesce multi makes it mutable, that's enough?

I'm fine with coalesce_single having a a slightly different signature, where the subquery is an immutable reference. As it is, it already has a different signature (accepting a single subquery instead of a vector of subqueries), and seems unnecessary change it.

rust-src/src/gridstore/coalesce.rs
Nmargolis: Tried this, but this turns matched into a reference, and doesn't consume coalesced, so context is then borrowed and can't be added to contexts.

PR #21, opened 2019-05-28
Coalesce tests

Nmargolis:
This PR adds more test coverage to coalesce, and is a continuation from #20 

### Summary of changes
- [x] Test matching for multiple languages
- [x] More tests for coalesce multi

### Next steps
- [ ] Continue adding tests
- [ ] Clean up tests for adjust_to_zoom
- [x] Move most of the tests in mod.rs to a proper test directory, and create structure for common test utils?


tests/coalesce_test.rs
```diff
@@ -0,0 +1,1259 @@
...
+        store: &store,
+        weight: 1.,
+        match_key: MatchKey { match_phrase: MatchPhrase::Range { start: 1, end: 3 }, lang_set: 1 },
+        idx: 1,
+        zoom: 14,
+        mask: 1 << 0,
+    };
+    let stack = vec![subquery];
+
+    println!("Coalesce single - NE proximity");
```
apendleton: Ideally we wouldn't print from within tests. If the goal is to split a test up into sections, it might be better to split them out into separate test functions. If multiple test functions need to share data or a built structure, consider the `lazy-static` pattern for sharing test data that we use in fuzzy-phrase: https://github.com/mapbox/fuzzy-phrase/blob/master/src/glue/mod.rs#L952-L998

tests/coalesce_test.rs
```diff
@@ -0,0 +1,1259 @@
...
+
+    // Test bbox at lower zoom than either of the expected results
+    println!("Coalesce multi - bbox at lower zoom than either of the expected results");
+    let stack = vec![
+        PhrasematchSubquery {
+            store: &store2,
+            weight: 0.5,
+            match_key: MatchKey {
+                match_phrase: MatchPhrase::Range { start: 1, end: 3 },
+                lang_set: u128::max_value(),
```
apendleton: we could think about aliasing `u128::max_value()` as `ALL_LANGUAGES` or something, just to make the intent here a little more clear

tests/coalesce_test.rs
Nmargolis: Definitely! After looking into this, there were some issues with Borrow and other traits not being implemented for the types generated by the lazy_static macro, so instead we could use once_cell Lazy, which doesn't use a macro and so is easier to understand how to implement: https://docs.rs/once_cell/0.2.1/once_cell/#lazy-initialized-global-data. Created a ticket for this: https://github.com/mapbox/carmen-core/issues/31

Nmargolis:
In the interest of getting this merged in so that the file structure can be reorganized, I have created tickets for all of the feedback I haven't gotten to yet.

- Utility functions: #30 
- Compiler warnings for unused variables: #29 (The optimization will use these variables)
- Use lazy static variables and smaller test functions instead of printing in tests #31

PR #23, opened 2019-05-31
Failure crate + Rust 2018 edition

apendleton:
This PR makes two tidying-up changes -- it's a tangent on the way to writing renumbering support in preparation for Carmen integration.

## failure
The first is to change how we're doing error handling to use the [failure](https://github.com/rust-lang-nursery/failure) crate. The `Result` type requires that you specify what types you're going to make your errors, and if you want to be able to use the `?` operator on results you get from calling other people's functions, you need to be able to coerce from whatever kind of errors you get back into the kind of error you return. The "grown up" way is for you to create your own error type, and then write converters from any kinds of errors you get into your error type. The laziest way is what we were doing before, which is `Box<dyn Error>` -- any kind of error, plus some vtable nonsense for good measure. `failure` is a library that's supposed to split the difference, making it practical to write your own error type if you need and mix that in with other people's errors without much boilerplate. (It's the successor to `error-chain`, which also serves about this purpose and is used, among other places, in neon-serde). I figured I would see how much work it was to switch, and it turns out... not much. Which is good, because I need to make my own error type for the renumbering thing I'm doing next.

## Rust 2018
While switching things over, I noticed in the `native` code for neon that had to add `extern crate` statements, which I didn't need to do in the main code. Turns out the main crate is using the [2018 edition](https://doc.rust-lang.org/edition-guide/rust-2018/index.html) of Rust, which doesn't require these anymore, but the Neon code was using 2015, so I switched it over for consistency.

@aarthykc turns out that also enabled [non-lexical lifetimes](https://doc.rust-lang.org/edition-guide/rust-2018/ownership-and-lifetimes/non-lexical-lifetimes.html), which means a lot of that boilerplate with it complaining about too many borrows of `&mut cx` fixed itself, and I was able to get rid of a bunch of those extra blocks and variables we had to add in to get `insert` to build before. So that's a bit tidier, too.

### DIFF
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -56,21 +56,21 @@ declare_types! {
             let grid_key = cx.argument::<JsObject>(0)?;
             let grid_entry = cx.argument::<JsValue>(1)?;
             let values: Vec<GridEntry> = neon_serde::from_value(&mut cx, grid_entry)?;
-            let js_phrase_id = {
-                grid_key.get(&mut cx, "phrase_id")?
-            };
-            let phrase_id: u32 = {
-                js_phrase_id.downcast::<JsNumber>().or_throw(&mut cx)?.value() as u32
-            };
-            let js_lang_set = {
-                grid_key.get(&mut cx, "lang_set")?
-            };
+            let phrase_id: u32 = grid_key
+                .get(&mut cx, "phrase_id")?
+                .downcast::<JsNumber>()
+                .or_throw(&mut cx)?
+                .value() as u32;
 
-            let js_array_lang = {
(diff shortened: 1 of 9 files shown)


native/src/lib.rs
```diff
@@ -51,37 +48,37 @@ declare_types! {
...
-            let js_lang_set = {
-                grid_key.get(&mut cx, "lang_set")?
-            };
-
-            let js_array_lang = {
-                js_lang_set.downcast::<JsArray>().or_throw(&mut cx)?
-            };
-
-            let lang_set: u128 = langarray_to_langset(&mut cx, js_array_lang)?;
+            let phrase_id: u32 = grid_key
```
aarthykc: YAY! SO much better!!

PR #24, opened 2019-05-31
Renumbering

apendleton:
Unlike carmen-cache, entries in the carmen-core grid store are keyed by phrase ID instead of by string. IDs will be assigned by fuzzy-phrase as of https://github.com/mapbox/fuzzy-phrase/pull/71 . However, the actual ID for each phrase isn't known until all phrases have been indexed, so in an indexing context in carmen, where we're processing features one at a time and adding their phrase and grid data to both the fuzzy-phrase index and the grid store, we need a temporary identifier. The fuzzy-phrase PR provides for returning a temporary ID after each fuzzy-phrase insert, and then for providing a mapping from temporary IDs to final IDs after the structure is built. This PR introduces a mechanism for consuming that mapping (expressed just as a vector of IDs, where the index is the temporary ID and the value at that index is the final ID), and under the assumption that all inserts so far have been with temporary IDs; it rewrites its internal temporary storage to use the new IDs instead of the old ones.

Also included is a Neon wrapper, which assumes the vector of u32's will be represented in Javascript as a JsArrayBuffer -- this matches the output of the corresponding function on the fuzzy-phrase side.

Tests are in place for Rust. I'll write some JS tests as well, but will wait until @aarthykc 's test work lands to avoid merge conflict unpleasantness.

### DIFF
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -96,6 +96,39 @@ declare_types! {
             }
         }
 
+        method renumber(mut cx) {
+            let js_id_map = cx.argument::<JsArrayBuffer>(0)?;
+            let mut this = cx.this();
+
+            let result: Result<(), String> = {
+                let lock = cx.lock();
+
+                let borrow_result = match js_id_map.try_borrow(&lock) {
+                    Ok(data) => {
+                        let slice = data.as_slice::<u32>();
+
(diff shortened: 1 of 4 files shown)


apendleton:
Update: we're going to plan on merging this without JS tests for now, to unblock other work, and add JS tests in a subsequent PR.

PR #33, opened 2019-06-13
[wip] Check for input validation for coalesce tests

aarthykc:
## Context
We tested most of the actual functionality of coalesce via #21, this pr adds testing the input arguments to coalesce to see if we are catching cases where we pass the wrong argument type and when the stack that we're sending to coalesce is invalid. 

## Next Steps 
- [ ] Review
- [ ] Merge

cc @apendleton @Nmargolis


aarthykc:
@apendleton added valid stacks and moved tests over to `tests` and removed the `test` directory.

PR #34, opened 2019-06-18
Ignore langarray values that are >= 128

aarthykc:
### Context 

While debugging one of the [failing tests in the integration of carmen-core with carmen](https://github.com/mapbox/carmen/blob/f93439c82c41de412dd4240dc0fef26437f33d87/test/acceptance/geocode-unit.konglish.test.js), it looked like the languages being set for the neighborhood index was `default` and `128`. `128` is what we set for [unmatched languages](https://github.com/mapbox/carmen/blob/master/index.js#L222) currently. Currently, the way we get by this in [carmen-cache](https://github.com/mapbox/carmen-cache/blob/master/src/node_util.hpp#L39) is to ignore any languages set to 128 or greater (since in rust and C++ the value gets wrapped to 1). This PR adds the same logic for `langarray_to_lang_set()`.

cc @Nmargolis, @apendleton

### DIFF
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -277,11 +277,12 @@ where
     if let Ok(lang_array) = maybe_lang_array.downcast::<JsArray>() {
         let mut out = 0u128;
         for i in 0..lang_array.len() {
-            if lang_array.get(cx, i)?.downcast::<JsNumber>().or_throw(cx)?.value() >= 128.0 {
+            let converted_lang_array = lang_array.get(cx, i)?.downcast::<JsNumber>().or_throw(cx)?.value();
+            if  converted_lang_array >= 128.0 {
                 continue;
             } else {
             out = out
-                | (1 << lang_array.get(cx, i)?.downcast::<JsNumber>().or_throw(cx)?.value() as usize);
+                | (1 << converted_lang_array as usize);
             }
         }
         Ok(out)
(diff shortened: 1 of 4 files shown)


native/src/lib.rs
```diff
@@ -277,8 +277,12 @@ where
     if let Ok(lang_array) = maybe_lang_array.downcast::<JsArray>() {
         let mut out = 0u128;
         for i in 0..lang_array.len() {
+            if lang_array.get(cx, i)?.downcast::<JsNumber>().or_throw(cx)?.value() >= 128.0 {
```
apendleton: This does the conversion from JS number to native number twice, once to do this check, and again to actually preserve the value if it succeeds. We should only do it once. So instead of (pseudocode):

```
if expensive_number_convert(number) < 128 {
    out = out | (1 << expensive_number_convert(number))
}
```

we should do:

```
converted = expensive_convert_number(number)
if converted < 128 {
    out = out | converted
}
```

aarthykc: That's fair!

native/src/lib.rs
```diff
@@ -277,8 +277,13 @@ where
     if let Ok(lang_array) = maybe_lang_array.downcast::<JsArray>() {
         let mut out = 0u128;
         for i in 0..lang_array.len() {
+            let converted_lang_array = lang_array.get(cx, i)?.downcast::<JsNumber>().or_throw(cx)?.value();
```
apendleton: Super nitpick, but I would convert to usize here. Comparing to `128.0` on the next line doesn't really make sense, semantically -- this isn't a floating point value, conceptually, it's an integer. But not a big deal either way.

Issue #3, opened 2019-04-04
Benchmarks


apendleton:
We should build out the machinery for Rust-side unit tests and benchmarks (the latter probably using `criterion`).

miccolis:
Capturing some discussion from chat w. @apendleton about benchmarks;

> The other option, though, would be to try and come up with a strategy for benching that actually at least sort of reflects production... on the fuzzy phrase side we have production-sized phrase dictionaries derived from open data sources, so maybe it's worth doing something like that... or, alternatively, building a benching tool that depends on private data, and then assembling a private corpus from our existing data sources
> In the past when benching coalesce and wanting to pass around a self-contained thing, I've used the `gb-*` indexes, which are big enough to be nontrivially but not so big that they can't be downloaded fairly quickly
> so if we went that route, we could maybe extract all the grids from current carmen cache into some easier-to-read format, put them on s3, and build some infrastructure to, during a bench run, download that data and index it, and run a bunch of real coalesces against it -- last time, I captured a bunch of actual coalesce calls from gb test runs

Nmargolis:
Added in #36, next steps outlined in #39

PR #36, opened 2019-06-19
Benchmarks

Nmargolis:
### Context
This PR adds some basic benchmark infrastructure.
Issue #3.

### Summary of changes
- [x] Adds Criterion benchmarks for Coalesce
- [x] Adds test util for loading grids from a line-delimited json file

### Next steps (future)
- [ ] Download a larger, more realistic dataset from s3 or other location
- [ ] Potentially add a [script](https://medium.com/@yamafaktory/rust-benchmarking-with-criterion-on-travis-ci-%EF%B8%8F-8b54d321e05) to run these benchmarks against master on CI


benches/coalesce.rs
```diff
@@ -0,0 +1,238 @@
...
+
+use carmen_core::gridstore::*;
+use test_utils::*;
+
+pub fn benchmark(c: &mut Criterion) {
+    // Make a vector to fill with closures to bench-test
+    let mut to_bench = Vec::new();
+
+    // Load data for coalesce single from json into a store
+    let filepath = Path::new("benches/data/coalesce-bench-single-3848571113.json");
```
apendleton: What is this relative to? Does it matter where you run `cargo bench` from?

rust-src/src/gridstore/store.rs
```diff
@@ -246,9 +246,6 @@ impl GridStore {
                     let coords_vec =
                         get_vector::<Coord>(record_ref.1, &rs_obj._tab, RelevScore::VT_COORDS)
                             .unwrap();
-                    // TODO could this be a reference? The compiler was saying:
-                    // "cannot move out of captured variable in an `FnMut` closure"
-                    // "help: consider borrowing here: `&match_opts`rustc(E0507)""
```
apendleton: Why remove these comments? (Mostly asking because this PR doesn't otherwise change this file, and it's going to cause a merge conflict with #37)

benches/coalesce.rs
Nmargolis: It's relative to the output of `std::env::current_dir()` which I think is the absolute path to the `carmen-core` dir. This seems to work regardless of whether the bench command is run in `carmen-core/` or `carmen-core/rust-src/` or `carmen-core/benches`, which suggests it's not dependent on where you run the bench command from. I added a comment to the load_grids_from_json function that the path should be relative to `carmen-core/`

rust-src/src/gridstore/store.rs
Nmargolis: Just because they seem like they shouldn't be in there. Agreed that it's not relevant to this PR, so reverting. At some point we can do a separate sweep for TODOs that probably don't belong in the code.

Nmargolis:
Added a comment about what the file path should be relative to, and reverted the removal of TODOs in an unrelated file.

The concern about using the same case over and over is valid and we should definitely have a more realistic set of queries to run. For now, this is essentially a port of the similarly-simplistic [carmen-cache bench](https://github.com/mapbox/carmen-cache/blob/master/bench/coalesce.bench.test.js)

PR #37, opened 2019-06-24
get_matching order change

apendleton:
This PR changes the output order of get_matching, particularly with respect to proximity. The old way:

* bucket everything by relev and score
* lazy-merge the buckets together by z-order coordinate
* for each relev + score + z-order, eagerly sort the ids

New way:

* bucket everything by relevance
* within each relevance, for each score, lazily map the coord stream to one that's annotated with scoredist (so, de-interlace the Morton encoding and calculate a real distance from the prox point)
* merge all the lazy-mapped streams for the given relevance by, first, (matches_language || within_radius), then by scoredist
* within each relev + (matches_language || within_radius) + scoredist, eagerly sort the ids

The big takeaways:
* we'll do fewer k-way merge operations (one per relevance instead of one per relevance + score), but each will be more expensive because they'll be merging from more sources
* we'll calculate the language-mismatch penalty, distance, and scoredist inside get_matching instead of inside coalesce

There are also a bunch of test updates, some warnings cleanup, and the odd miscellaneous structural tweak or two.


rust-src/src/gridstore/coalesce.rs
```diff
@@ -190,8 +167,7 @@ fn coalesce_multi<T: Borrow<GridStore> + Clone>(
         let grids =
             subquery.store.borrow().get_matching(&subquery.match_key, &adjusted_match_opts)?;
 
-        // TODO: limit how many grids we consume
-        for grid in grids {
+        for grid in grids.take(100_000) {
```
Nmargolis: Why 100_000? Might help to have an explanation here of why this number.

rust-src/src/gridstore/coalesce.rs
```diff
@@ -301,7 +277,7 @@ fn coalesce_multi<T: Borrow<GridStore> + Clone>(
 }
 
 /// Calculates the tile distance between a proximity x and y and a grid x and y
-fn tile_dist(proximity_x: u16, proximity_y: u16, grid_x: u16, grid_y: u16) -> f64 {
+pub fn tile_dist(proximity_x: u16, proximity_y: u16, grid_x: u16, grid_y: u16) -> f64 {
```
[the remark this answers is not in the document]
apendleton: Moved 👍

rust-src/src/gridstore/coalesce.rs
apendleton: Added a comment

apendleton:
@Nmargolis should be ready for re-review

PR #40, opened 2019-07-04
Benchmarks and performance improvements

Nmargolis:
This PR has several chunks of work.

First, benchmark-related utilities
- [x] introduce a serialization mechanism for dumping the contents of a gridstore RocksDB database, and for creating a new database from dumped data
- [x] introduce some utility functions for loading compressed benchmark data from S3
- [x] shift around the existing coalesce benchmarks so they still work even though they differ in data format from the new ones
- [x] add a script for profiling bench runs

Then, actual benchmarks:
- [x] add real-data-driven benchmarks for coalesce with and without autocomplete, and with and without proximity, of varying data types

Then, the first round of changes driven by the benchmarks:
- [x] enable RocksDB compression (lz4hc)
- [x] enable mmap RocksDB reads

Unrelated to the above:
- [x] it turns out `cargo kcov` rebuilds all of our tests with different compiler flags, and prints no output while it does so... because of the new dev dependencies around AWS/S3 introduced for the above, builds now take longer than before, causing this build step to take more than 10 minutes, which stalls Travis and causes our job to be killed. Two changes related to that: I split out coverage into a separate Travis build worker, and wrapped the actual `cargo kcov` step in a node script that periodically prints things to keep Travis from dying. (I also tried switching us back to tarpaulin, but we appear to still be affected by this segfault issue there.)


keepalive.js
```diff
@@ -0,0 +1,15 @@
+const { spawn } = require('child_process');
```
aarthykc: This is to get Travis to pass, right?

Issue #38, opened 2019-06-24
Prefix store


apendleton:
Carmen-cache indexes grid contents multiple times so as to pre-combine prefixes to make autocomplete not take a million years. We'll want to do something similar here, but we don't actually know what letters are in the phrase we're looking for, so it'll be a somewhat more complicated operation.

In carmen-cache, naively if we're doing a prefix scan, we end up zipper-merging grid entries for each phrase that starts with that prefix, so if you search for everything that starts with "a," we'd get:
```
a => grid, grid, grid, ...
aardvark => grid, grid, grid, ...
aardwolf => grid, grid, grid, ...
abacate => grid, grid, grid, ...
abacinate => grid, grid, grid, ...
... etc.
```

to make that merger simpler, at index time, we loop through all the entries in the index and re-index them pre-cached by prefix, at two different lengths. We have special entries for every three-character prefix that exists in the index, and every six-character prefix that exists in the index, with all the grid entries for all of the phrases that start with those prefixes precombined. We mark those cached pre-sorted lists with `=1` and `=2`, respectively. So if you search for "aba," rather than scanning all words that start with "aba," we'll look up `=1aba` and read from that single entry instead. If you look for just "a," we'll scan everything that starts with `=1a` instead of `a` so that at least some of the sorting work is already done. Likewise, if you search for "abac," we'll look for everything that starts with `=2abac` to take advantage of the six-character precombined list.

Here, we won't know what the actual words and phrases we're indexing are, since all phrases will have been assigned a number. Instead, we'll want to group them numerically. For example, we might choose to precombine groups of 1000 entries. So we'll have phrases 0, 1, 2, 3, 4, 5, etc., and generate pre-combined groups for 0-999, 1000-1999, 2000-2999, etc. If a user searches for a prefix range 998-3055, we'll do three scans: 998-999 (regular entries), 1xxx-2xxx (two pre-combined entries), and 3000-3055 (regular entries).

Each grid key in the grid store has a first byte that's a "type marker." This byte is presently always 0. We can use a different value here to indicate a pre-combined list -- perhaps 1. We'll also need to decide how big each chunk of precombined entires should be. It might be good to look at the actual contents of some real, large carmen-cache indexes to get a good sense of this. We'll probably actually want to do them in power-of-two sizes rather than power-of-10 sizes as per my example (so maybe chunks of 1024 instead of chunks of 1000 or whatever) to make the math for deciding what bin a number goes into as fast as possible.

[...]

PR #41, opened 2019-07-16
[WIP] Prefix Store

aarthykc:
### Builder
Per #38, this PR creates a prefix store and modifies `get_matching()` to read from the prefix store. When we call `finish()` the idea is that we create pre combined groups of 1024 entries. For example, grid entries with `phrase_ids` from 0 - 1023 will be combined under one `phrase_id`.

### Reader
In phrasematch when we get a range of grids to look up, we check to see if the range starts before, at or after the prefix bucket (since they're buckets of 1024, prefix ranges will be 1024 - 2047 etc) and similarly for the end. If the range starts before the prefix we'll iterate through the regular entries, then the prefix entries all the way to the end of the range provided. For example, if the user searches for the range 1000 - 2049 we'll scan through the following: 
- 1000 - 1023 - Regular entries 
- 1024 - 2047 - Prefix entries 
- 2048 - 2049 - Regular entries

## Next actions:
- [x] Publish a dev version and see if tests still pass for carmen - [Yes the do!](https://github.com/mapbox/carmen/commits/prefix-store)
- [x] Get the PR reviewed 
- [x] Merge

cc @apendleton @Nmargolis

### DIFF
--- a/rust-src/src/gridstore/builder.rs
+++ b/rust-src/src/gridstore/builder.rs
@@ -70,7 +70,7 @@ fn get_fb_value(value: &mut BuilderEntry) -> Result<Vec<u8>, Error> {
     let record =
         PhraseRecord::create(&mut fb_builder, &PhraseRecordArgs { relev_scores: Some(fb_rses) });
     fb_builder.finish(record, None);
-    Ok(fb_builder.finished_data().to_owned())
+    Ok(fb_builder.finished_data().to_vec())
 }
 
 impl GridStoreBuilder {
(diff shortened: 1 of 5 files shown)


rust-src/src/gridstore/builder.rs
```diff
@@ -34,6 +34,45 @@ fn extend_entries(builder_entry: &mut BuilderEntry, values: &[GridEntry]) -> ()
...
+            &mut fb_builder,
+            &RelevScoreArgs { relev_score: *rs, coords: Some(fb_coords) },
+        );
+        rses.push(fb_rs);
+    }
+    let fb_rses = fb_builder.create_vector(&rses);
+    let record =
+        PhraseRecord::create(&mut fb_builder, &PhraseRecordArgs { relev_scores: Some(fb_rses) });
+    fb_builder.finish(record, None);
+    Ok(fb_builder.finished_data().to_owned())
```
aarthykc: Note to self: remove `to_owned()`

PR #43, opened 2019-07-30
Fix proximity radius comparison and use strict dedupe

Nmargolis:
### Proximity radius check for language penalty
A language penalty is applied to the relevance score of features that don't match the language of the subquery. When proximity is specified, the language penalty is not applied if the grid is within the proximity radius. This provides better results for the  "tourist" use-case, so nearby results are  surfaced even if the language is set to a non-local language, instead of preferring features far away that happen to match the language.

The current implementation compares the distance from the proximity points (in tiles) to the raw proximity radius in the match opts (in miles).

This PR updates this comparison to compare apples to apples by converting the proximity radius to tiles. It is slightly duplicative, since the proximity radius also gets converted to tiles in the scoredist function, so there may be opportunities to optimize by only doing the calculation once in the future.

### Real deduping by feature id
The current implementation uses rust's `dedupe_by_key` method in coalesce to "dedupe" contexts with the same grid id. This method requires the vector to be already sorted by the key to properly remove all duplicates, otherwise it only removed duplicates that happen to be next to each other in the vector. An approach similar to this worked in carmen-cache, but in carmen-core, contexts get sorted by relev, scoredist, etc, so it's possible for there to be many grids with the same id in the first several entries, without being next to each other. This was visible in a [carmen test](https://github.com/mapbox/carmen/blob/master/test/acceptance/geocode-unit.proximity-polygon.test.js) for behavior of features that span many tiles. The test indexes 3 features, each spanning several hundred tiles, and expects all three features to be returned in the results. Previously, only two were returned because the first context that had the 3rd grid id was well past the 40 feature cutoff.

This PR uses a hashmap in coalesce single to keep track of grid ids that have already been seen, and replaces the value associated with that id only if a coalesce entry has a higher scoredist and higher or equal relevance.

## Summary of changes
- [x] Convert proximity radius to tiles to properly compare with the distance
- [x] Use a hashmap to keep track of feature ids that have been added to avoid duplicates instead of using `dedupe_by_key`

## Next steps
- [ ] Evaluate whether it's worth a refactor to only do the proximity radius calculation once
- [ ] Assess whether a similar dedupe strategy is necessary for coalesce multi

### DIFF
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -1,5 +1,6 @@
 use std::borrow::Borrow;
 use std::cmp::Reverse;
+use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 
 use failure::Error;
@@ -110,16 +111,16 @@ fn coalesce_single<T: Borrow<GridStore> + Clone>(
         let current_scoredist = coalesce_entry.scoredist;
 
         // If it's the same feature as one that's been added before, but a higher scoredist, update the entry
-        match coalesced.get(&current_id) {
-            Some(already_coalesced) => {
-                if current_scoredist > already_coalesced.scoredist
-                    && current_relev >= already_coalesced.grid_entry.relev
+        match coalesced.entry(current_id) {
+            Entry::Occupied(mut already_coalesced) => {
+                if current_scoredist > already_coalesced.get().scoredist
+                    && current_relev >= already_coalesced.get().grid_entry.relev
                 {
-                    coalesced.insert(current_id, coalesce_entry);
+                    already_coalesced.insert(coalesce_entry);
                 }
             }
-            None => {
-                coalesced.insert(current_id, coalesce_entry);
+            Entry::Vacant(entry) => {
+                entry.insert(coalesce_entry);
             }
         }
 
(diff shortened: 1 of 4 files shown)


rust-src/src/gridstore/coalesce.rs
```diff
@@ -102,27 +103,49 @@ fn coalesce_single<T: Borrow<GridStore> + Clone>(
...
+        let current_relev = coalesce_entry.grid_entry.relev;
+        let current_scoredist = coalesce_entry.scoredist;
+
+        // If it's the same feature as one that's been added before, but a higher scoredist, update the entry
+        match coalesced.get(&current_id) {
+            Some(already_coalesced) => {
+                if current_scoredist > already_coalesced.scoredist
+                    && current_relev >= already_coalesced.grid_entry.relev
+                {
+                    coalesced.insert(current_id, coalesce_entry);
```
apendleton: Ideally, instead of doing a `get` and then an `insert` here, we'd use the `entry` API: https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry . `get` and `insert` do duplicated work in that they both hash the key and figure out whether or not it's already in the map. The `entry` API returns a result that can both tell you if the key is already present (and what its current value is) and also let you mutate that entry in the hashtable. I think you can do what you want here (handling both the key already being there and not) via the `and_modify`/`or_insert` pattern seen here: https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.and_modify

This can also wait for a followup PR though -- it's a performance optimization, but probably not pressing.

Nmargolis: ooh I had looked into that but didn't realize there was the `and_modify` method. It's a little more complicated than just insert if it's not there, since we also want to do the check for if the scoredist is higher or not. I think we could do that in the `and_modify` closure.

PR #44, opened 2019-08-13
Lightweight append

apendleton:
This PR introduces an alternative interface to the `append` mechanism used in index building. It takes advantage of the fact that in practice, Carmen's indexer indexes chunks of many grid entries that for the same feature that are the same on all dimensions (relev, score, id, hash) except location. The new version avoids all the complex-object deserialization (and instantiation of tons of JS strings needed to read from different fields on a JS object) by accepting all the non-changing features for each chunk as bare arguments, and the X and Y coordinates, which do vary, as a simple array of arrays of integers.

**THIS NEW FUNCTION HAS NO TESTS.**


apendleton:
Closing in favor of #60

PR #45, opened 2019-08-13
Reuse ids

apendleton:
This PR changes the Flatbuffers format for carmen-core to take advantage of the fact that it's often the case, particular with polygonal features, for many X/Y entries for the same phrase to map to the same ID or list of IDs. It hoists these lists of IDs up to a unique-ified top-level list (so, a list of lists of IDs), and then changes the format for the object that *was* `{x, y, [id]}` to be `{x, y, id_idx}`, where the last field is an index into the aforementioned list of lists. This both lets us dedupe these ID lists, and lets us store this x/y/ids object as a Flatbuffers struct instead of a table, which eliminates an extra layer of indirection.


apendleton:
Closing as stale -- we moved away from flatbuffers altogether (though the new format reuses this idea)

PR #46, opened 2019-08-28
Variable prefix bins [wip]

apendleton:
#41 introduced a mechanism for pre-combining the grids corresponding to ranges of phrase IDs so that autocomplete scans of large ranges of IDs could combine their contents more efficiently. It turns out in practice that the strategy introduced there was too slow. It created fixed-size bins of ID ranges to precombine (1024 phrases apiece), and would assemble sets of buffers to read for an autocomplete query by scanning individual phrase entries up until the next precombined bin start boundary, then reading from the precombined bins for awhile until the end boundary of the last precombined bin that fit fully within the sought range, then reading individual phrases again until the end of the sought range.

This meant potentially lots of individual phrase reads, as many as three different database seeks per get_matching call, and the existence of at least some bins that could never be used in practice, because they might contain multiple phrases that didn't actually contain any real textual prefixes. The whole thing seems to be a regression from carmen-cache where bin boundaries correspond tightly with actual textual query prefixes, but unlike carmen-cache, carmen-core doesn't know what actual text corresponds to each phrase ID.

This branch ditches the fixed-size bins in favor of a mechanism to allow external callers to specify custom bin boundaries (that presumably correspond with actual text prefixes). These boundaries are then used to generate variably sized bins, and are then stored in a special key in the database for later retrieval, and are then retrieved at index load time and used at query time. There are some benchmarks as well with manually specified boundaries for our sample datasets.

Note: the mechanism for actually splitting a range of phrases into sensibly sized bins is not specified here, and will be up to outside callers to implement.

### DIFF
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -108,7 +108,7 @@ declare_types! {
             }
         }
 
-        method compact_append(mut cx) {
+        method compactAppend(mut cx) {
             let grid_key = cx.argument::<JsObject>(0)?;
             let phrase_id: u32 = grid_key
                 .get(&mut cx, "phrase_id")?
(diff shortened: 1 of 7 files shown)


apendleton:
Closing in favor of #60, which incorporates this work.

PR #50, opened 2020-01-06
Tests for compactAppend and clean up 

aarthykc:
## Context 
As a part of getting better test coverage on carmen-core after a number of functions were added to greatly improve query time and indexing speed, this PR adds tests for `compact_append()`, `compactApend()` functions. This PR also annotates as well as removes functions that are not being used anywhere and formats all the files to get tests to pass.

cc @apendleton @CyanRook @Nmargolis

### DIFF
--- a/native/Cargo.lock
+++ b/native/Cargo.lock
@@ -70,22 +70,23 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.47.3"
+version = "0.49.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "cexpr 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


rust-src/src/gridstore/gridstore_format.rs
```diff
@@ -42,29 +43,19 @@ impl<T: UniformEncodable> UniformScalarOffset<T> {
...
-    phantom: PhantomData<T>
+    phantom: PhantomData<T>,
 }
 
 impl<T: FixedEncodable> FixedVecOffset<T> {
     fn new(addr: usize) -> Self {
         Self { addr, phantom: PhantomData }
     }
-
-    fn from_fixed_pointer(data: &[u8], offset: usize) -> Self {
```
CyanRook: This is interesting, we need FixedVecOffset, but don't use these two helper functions, even though they match what other encodings look like.  I think we might want to keep these to keep the encodings consistent but might want to put it in a trait that requires implementation to avoid dead code warnings.

CyanRook: Possibly something along the lines of FixedDerivable/VarDecodable or something, I'm not sure what to call it.

PR #54, opened 2020-01-08
spatial.rs Cleanup

CyanRook:
* Remove TODOs
* Support all zoom levels

### DIFF
--- a/rust-src/src/gridstore/spatial.rs
+++ b/rust-src/src/gridstore/spatial.rs
@@ -1,7 +1,9 @@
 use crate::gridstore::gridstore_format::{Coord, UniformVec};
 use itertools::Itertools;
 use morton::{deinterleave_morton, interleave_morton};
+use std::cmp::max;
 use std::cmp::Ordering::{Equal, Greater, Less};
+use std::collections::HashMap;
 
 /// Generate a tuple of the (min, max) range of the Coord Vector that overlaps with the bounding box
 ///
@@ -60,7 +62,7 @@ pub fn bbox_filter<'a>(
     let range = bbox_range(coords, bbox)?;
     Some((range.0..=range.1).filter_map(move |idx| {
         let grid = coords.get(idx as usize);
-        let (x, y) = deinterleave_morton(grid.coord); // TODO capture this so we don't have to do it again.
+        let (x, y) = deinterleave_morton(grid.coord);
         if x >= bbox[0] && x <= bbox[2] && y >= bbox[1] && y <= bbox[3] {
             return Some(coords.get(idx as usize));
         }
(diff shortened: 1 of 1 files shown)


rust-src/src/gridstore/spatial.rs
```diff
@@ -1,7 +1,9 @@
 use crate::gridstore::gridstore_format::{Coord, UniformVec};
 use itertools::Itertools;
 use morton::{deinterleave_morton, interleave_morton};
+use std::cmp::max;
 use std::cmp::Ordering::{Equal, Greater, Less};
+use std::collections::HashMap;
```
CyanRook: Remove unused

rust-src/src/gridstore/spatial.rs
```diff
@@ -462,7 +464,7 @@ fn tile_dist_test() {
 }
 
 /// Returns the number of tiles per mile for a given zoom level
-const fn tiles_per_mile_by_zoom(zoom: u16) -> f64 {
+fn tiles_per_mile_by_zoom(zoom: u16) -> f64 {
```
CyanRook: I'm not sure why we need this as a constant function over a regular function.

rust-src/src/gridstore/spatial.rs
```diff
@@ -513,7 +519,7 @@ fn tiles_per_mile_by_zoom_test() {
 /// Convert proximity radius from miles into scaled number of tiles
 #[inline]
 pub fn proximity_radius(zoom: u16, radius: f64) -> f64 {
-    debug_assert!(zoom <= 16);
+    //debug_assert!(zoom <= 16);
```
CyanRook: Remove or uncomment

rust-src/src/gridstore/spatial.rs
```diff
@@ -60,7 +62,7 @@ pub fn bbox_filter<'a>(
     let range = bbox_range(coords, bbox)?;
     Some((range.0..=range.1).filter_map(move |idx| {
         let grid = coords.get(idx as usize);
-        let (x, y) = deinterleave_morton(grid.coord); // TODO capture this so we don't have to do it again.
+        let (x, y) = deinterleave_morton(grid.coord);
```
CyanRook: This operation is fast.  We might be able to find a way to push it through, but the function is inlined and only consists of a half dozen bit operations with no branching.

rust-src/src/gridstore/spatial.rs
apendleton: We don't, but on the other hand, it doesn't hurt anything... declaring functions as `const` in Rust should just declare that they *can* be calculated at compile time assuming constant input (which is true in this case).

rust-src/src/gridstore/spatial.rs
apendleton: I guess remove?

rust-src/src/gridstore/spatial.rs
```diff
@@ -537,7 +543,7 @@ fn proximity_radius_test() {
         1.2485901539399482,
         "proximity_radius in tiles for zoom 6, radius 40 is as expected"
     );
-    // TODO: test zoom > 14?
+    assert_eq!(proximity_radius(17, 400.), 1080.0, "proximity_radius should work for zoom 14");
```
apendleton: should be zoom 17, no?

rust-src/src/gridstore/spatial.rs
CyanRook: I guess it depends on how we want to handle out of bounds zooms, since floating point operations are not const in stable

PR #56, opened 2020-01-09
Add more tests for the variable-bins/bin-boundaries logic

apendleton:
This adds more tests for the variable-bins logic, to make sure the behavior is consistent with indexes without pre-combined bins. It also adds tests in JS, of which there weren't any before.

One other big change: it alters the sort order for output from coalesce, to be sorted by `(relev desc, scoredist desc, x desc, y desc, id desc)` so that it's the same as things are ordered in storage, instead of what it was before and is in carmen-cache, which is `(relev desc, scoredist desc, id asc, x asc, y asc)`. The ordering by x/y/id is arbitrary and just for consistency, and there's no reason any particular ordering is better than any other ordering, but the fact that different parts of the code used different sort orders was causing some slightly inconsistent behaviors between the version of the code with bins and the version without. This might make slight differences in api-geocoder in cases where there are ties, but we shouldn't see changes that are either consistently better or worse, just different.

### DIFF
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -154,13 +154,13 @@ fn coalesce_single<T: Borrow<GridStore> + Clone>(
         .collect();
 
     contexts.sort_by_key(|context| {
-        (
-            Reverse(OrderedFloat(context.relev)),
-            Reverse(OrderedFloat(context.entries[0].scoredist)),
-            context.entries[0].grid_entry.id,
+        Reverse((
+            OrderedFloat(context.relev),
+            OrderedFloat(context.entries[0].scoredist),
             context.entries[0].grid_entry.x,
             context.entries[0].grid_entry.y,
-        )
+            context.entries[0].grid_entry.id,
+        ))
     });
 
     contexts.truncate(MAX_CONTEXTS);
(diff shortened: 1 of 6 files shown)


rust-src/src/gridstore/coalesce.rs
```diff
@@ -154,13 +154,13 @@ fn coalesce_single<T: Borrow<GridStore> + Clone>(
         .collect();
 
     contexts.sort_by_key(|context| {
-        (
-            Reverse(OrderedFloat(context.relev)),
-            Reverse(OrderedFloat(context.entries[0].scoredist)),
-            context.entries[0].grid_entry.id,
+        Reverse((
```
CyanRook: Nice

rust-src/src/gridstore/coalesce.rs
```diff
@@ -311,9 +311,9 @@ fn coalesce_multi<T: Borrow<GridStore> + Clone>(
             Reverse(OrderedFloat(context.relev)),
             Reverse(OrderedFloat(context.entries[0].scoredist)),
             context.entries[0].idx,
-            context.entries[0].grid_entry.id,
-            context.entries[0].grid_entry.x,
-            context.entries[0].grid_entry.y,
+            Reverse(context.entries[0].grid_entry.x),
```
CyanRook: Why does this not match the Reverse() of the change above?

rust-src/src/gridstore/mod.rs
```diff
@@ -569,6 +597,42 @@ mod tests {
                 scoredist: 1.0,
             })
         }
-        assert_eq!(records, expected);
+        assert_eq!(records_with_boundaries, expected);
```
CyanRook: This seems like 3 distinct test that are all in the same test because of similar setup. While it looks safe from side effects of one test affecting the other, generally it is best practice to split this up, even if it means test/ci will take a little bit longer.

rust-src/src/gridstore/coalesce.rs
apendleton: `idx` isn't in the other one, and I wanted to keep the ordering of `idx` the same -- we still want to prefer results from low-numbered indexes like country over high-numbered indexes like POIs, all things being equal. So mirroring the other one would have meant Reverse on the whole thing, and then Reverse *again* of this one, and that felt worse. 🤷‍♂️

PR #58, opened 2020-01-09
Spatial tests

apendleton:
With the switch to a homegrown data format, I had commented out a bunch of tests in `spatial.rs` that depended on constructing a `flatbuffers` record and then operating on it with spatial functions. This PR uncomments all of that, and ports it to use the new format.

Obscured in the giant diff are a few functions that have actually had their results change since they were commented out. I'll point them out in comments, but in a nutshell, in some cases the sort order for spatial sorts will have slightly changed in instances where multiple possible results were tied. Bisecting suggests that this happened with [this commit](https://github.com/mapbox/carmen-core/commit/59350ea3c1eee635e55c5a0a251d2ec92df43f19) back in September, which ported some uses of `kmerge_by` in the spatial code to `merge_by` for efficiency reasons; merge_by seems to occasionally make different (arbitrary) decisions in the case of ties. Seems fine though.

### DIFF
--- a/rust-src/src/gridstore/spatial.rs
+++ b/rust-src/src/gridstore/spatial.rs
@@ -3,6 +3,11 @@ use itertools::Itertools;
 use morton::{deinterleave_morton, interleave_morton};
 use std::cmp::Ordering::{Equal, Greater, Less};
 
+#[cfg(test)]
+use crate::gridstore::gridstore_format;
+#[cfg(test)]
+use crate::gridstore::common::relev_float_to_int;
+
 /// Generate a tuple of the (min, max) range of the Coord Vector that overlaps with the bounding box
 ///
 /// Returns (Some(min,max)) if the Coord Vector morton order range overlaps with the bounding box,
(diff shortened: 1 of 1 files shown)


rust-src/src/gridstore/spatial.rs
```diff
@@ -184,256 +189,273 @@ fn coord_binary_search<'a>(
...
+
+    #[test]
+    fn proximity_search() {
+        let buffer = encoded_val_generator((1..10).rev()); // [9,8,7,6,5,4,3,2,1]
+        let reader = gridstore_format::Reader::new(buffer.as_slice());
+        let coords = get_coords_from_reader(&reader);
+
+        let result = proximity(coords, [3, 0]).unwrap().map(|x| x.coord).collect::<Vec<u32>>();
+        assert_eq!(
+            vec![5, 4, 6, 3, 7, 2, 8, 1, 9],
```
apendleton: this one used to end 9, 1 -- but they're tied in terms of distance

rust-src/src/gridstore/spatial.rs
```diff
@@ -184,256 +189,273 @@ fn coord_binary_search<'a>(
...
+        let coords = get_coords_from_reader(&reader);
+        assert_eq!(proximity(coords, [3, 0]).is_none(), true);
+
+        let sparse: Vec<u32> = vec![24, 21, 13, 8, 7, 6, 1]; // 1 and 13 are at the same distance from 7
+        let buffer = encoded_val_generator(sparse.into_iter());
+        let reader = gridstore_format::Reader::new(buffer.as_slice());
+        let coords = get_coords_from_reader(&reader);
+        let result = proximity(coords, [3, 1]).unwrap().map(|x| x.coord).collect::<Vec<u32>>();
+        assert_eq!(
+            vec![7, 6, 8, 1, 13, 21, 24],
```
apendleton: in this one, the 1 and 13 swapped places as compared to the old version, but again, they're tied

PR #57, opened 2020-01-09
store.rs cleanup

CyanRook:
* Remove todo as won't do due to lifetime issues
* Add dyn explicitly due to rust best practices

PR #60, opened 2020-01-20
All the things to actually use carmen-core

apendleton:
This is a giant PR that has the accumulation of a whole bunch of work necessary to get carmen-core production-ready. Some parts have already been reviewed as PR's into this branch; I'll link to those below as appropriate.

# New data format
This branch ditches flatbuffers altogether in favor of a homegrown data format that's significantly more compact. It omits all in-format type-tagging or versioning, and most attempts at alignment, to shave off bytes. Additionally, it makes heavy use of variable-length encoding where possible. It's defined as a series of reusable traits for encoding and decoding that are reusable across types, and provides ways of encoding individual instances of a data type implementing these traits, or vectors of instances of these types.

Vectors come in three flavors: fixed, variable, and uniform. With fixed vectors, the size of a given type is always the same, and the size of the vector is just that size times the length of the vector. Random access is possible by multiplying an offset by the size. For variable vectors, the type must provide a way to write an instance using a variable-length encoding. Seek is not possible because not all elements are the same size, so elements can only be accessed by iterating from the beginning. Uniform vectors split the difference: at encoding time, a determination is made as to the maximum number of bytes needed to store any value in the vector, and that number of bytes is used to store all of the elements (so for example, for encoding u32 values, if a given vector's largest value only requires 3 bytes to encode, all elements will be encoded with 3 bytes). A one-byte tag is added at the start of the vector in addition to the length to represent this per-element size. This saves space as compared to fixed encoding, but less so than variable encoding; it does, however, preserve the ability to do random seeking (necessary for, e.g., binary search).

# compact appending
Carmen-cache used a single integer to encode all the contents of a "grid" (relev, score, x, y, id). Carmen-core initially moved to passing full objects back and forth instead, so each grid would be an object with each of those fields. The process of marshalling data out of those complex objects at indexing time turned out to be prohibitive, however. This branch adds a new method, `compact_append`, that's faster in many circumstances. It takes advantage of the fact that in typical use, many features will require many grids to be inserted, one for each tile coordinate, and all will be for the same phrase, and have the same relevance, score, ID, and phrase hash. Each compact-append call is for a single set of grids for one feature like this, so it can take a list of coordinate pairs, and one each of all the other values, and take them all as positional arguments rather than named attributes on an object (retrieving a named field requires instantiating a JS string from Rust and passing it into a Node-side method to fetch parameters by name). This is significantly faster than the standard append method.

# streaming get_matching
The get_matching implementation in master turned out to be too slow, so it's been replaced with a different algorithmic approach. The version in master waited as long as possible to begin decoding any given component of a grid store record. This meant when combining several records, it would first sort them by relevance, then for the best relevance, sort by score, then for the best score, sort by distance, etc. It wouldn't begin to retrieve an ID until all of that sorting was finished, and only do so from the best record. This was maximally lazy, which would hopefully have meant better speed due to less wasted work, but it turned out it meant we didn't know until we had finished the operation which records we'd actually end up retrieving values from, so we had to hold them all in memory until we were done. It also ended up structured as a complex series of nested N-way merge operations, which in practice seemed to have lots of overhead.

The new implementation takes each record and transforms it into a lazy iterator of its grids from best (by relevance, score/dist, etc.) and then uses a priority queue to merge these iterators together. This means slightly more work up front on each record, but ultimately involves fewer levels of nested merging. Additionally, in cases where we're only retrieving a small number of records (like the 80 we retrieve for coalesce-single), we can know up front that the most records we'll ever read from is 80 thanks to the pigeon-hole principle, so we can start with the 80 records whose best values are best, and throw the rest away right off the bat. This both saves significant memory and makes subsequent merge operations (i.e., mutations of the priority queue) much faster, as their complexity is relative to the total size of the queue. We use a min-max heap to make this work, so we can both retrieve values from the best iterator (the max) and discard elements from the other end (the min end).

# Variable bins

Carmen-core master's strategy for anticipating short autocomplete queries' needs to read from lots of records is to create fixed-size pre-combined bins by ID, without regard to whether or not the boundaries between bins actually correspond to the boundaries between prefixes. This turns out to be very slow in practice, and it works much better to have bin boundaries actually align with prefix boundaries. We add the ability to extract this information from fuzzy-phrase in mapbox/fuzzy-phrase#72 and this branch now allows bin boundaries to be loaded in at build time and then used to construct these bins during `finish()`, baked into the index, and used at runtime.

# Things that have already been reviewed on their way into this branch
* A fix for a major performance issue that caused runaway grid comparisons in some circumstances in coalesce: #51
* A return to tarpaulin for code coverage: #55 
* Improved test coverage and code cleanup: #50 #52 #53 #54 #56 #57 #58 #59


rust-src/src/gridstore/builder.rs
```diff
@@ -119,22 +172,36 @@ impl GridStoreBuilder {
...
     }
 
+    pub fn load_bin_boundaries(&mut self, bin_boundaries: Vec<u32>) -> Result<(), Error> {
+        self.bin_boundaries = bin_boundaries;
+        Ok(())
+    }
+
     /// Writes data to disk.
-    pub fn finish(mut self) -> Result<(), Error> {
+    pub fn finish(self) -> Result<(), Error> {
```
CyanRook: We should remove the comment and rename this function to write_to_disk

rust-src/src/gridstore/coalesce.rs
```diff
@@ -185,28 +194,34 @@ fn coalesce_multi<T: Borrow<GridStore> + Clone>(
...
-            subquery.store.borrow().get_matching(&subquery.match_key, &adjusted_match_opts)?;
+
+        if zoom_adjusted_match_options.zoom != subquery.zoom {
+            zoom_adjusted_match_options = match_opts.adjust_to_zoom(subquery.zoom);
+        }
+
+        let grids = subquery.store.borrow().streaming_get_matching(
+            &subquery.match_key,
+            &zoom_adjusted_match_options,
+            100_000,
```
CyanRook: This should be a labelled constant used here and below, not two bare numbers.

rust-src/src/gridstore/common.rs
```diff
@@ -161,7 +160,8 @@ impl MatchOpts {
                         // Scale the top left (min x and y) tile coordinates by 2^(zoom diff).
                         // Scale the bottom right (max x and y) tile coordinates by 2^(zoom diff),
                         // and add the new number of tiles (-1) to get the outer edge of possible tiles.
-                        // TODO comment explaining why parens for -1
+                        // We subtract 1 from the scale_multiplier before adding to prevent an integer overflow
+                        // given that we're using a 16bit integer
```
CyanRook: This might be more clear in the form
`(orig_bbox[2] + 1) * scale_multiplier - 1)`
This logic isn't specific to 16 bit integers, but to powers of 2 in general.

rust-src/src/gridstore/gridstore_format.rs
```diff
@@ -0,0 +1,567 @@
...
+
+    #[allow(dead_code)]
+    pub fn iter(&self) -> impl Iterator<Item = T> + '_ {
+        (0..self.len).map(move |idx| self.get(idx))
+    }
+
+    pub fn into_iter(self) -> impl Iterator<Item = T> + 'a {
+        (0..self.len).map(move |idx| self.get(idx))
+    }
+}
```
CyanRook: I feel like the above code is a data agnostic encoding mechanism and should be separated out from our domain specific code below.  This file is gigantic and contains both fundamental interface style code and implementation code in the same file.

CyanRook: We should probably have 1-3 files for encoding styles, a file for RelevanceScore, a file for Coordinates, and a file PhraseRecords

rust-src/src/gridstore/spatial.rs
```diff
@@ -125,21 +126,19 @@ pub fn bbox_proximity_filter<'a>(
...
-        Box::new((range.0..prox_mid).rev().filter_map(filtered_get)) as Box<Iterator<Item = Coord>>;
-    let tail =
-        Box::new((prox_mid..=range.1).filter_map(filtered_get)) as Box<Iterator<Item = Coord>>;
-    let coord_sets = vec![head, tail].into_iter().kmerge_by(move |a, b| {
-        let d1 = (a.coord() as i64 - prox_pt) as i64;
-        let d2 = (b.coord() as i64 - prox_pt) as i64;
+    let head = (range.0..prox_mid).rev().filter_map(filtered_get);
+    let tail = (prox_mid..=range.1).filter_map(filtered_get);
+    let coord_sets = head.into_iter().merge_by(tail.into_iter(), move |a, b| {
+        let d1 = (a.coord as i64 - prox_pt) as i64;
```
CyanRook: What is d1 d2? Can we be more descriptive?

rust-src/src/gridstore/spatial.rs
```diff
@@ -174,15 +173,15 @@ fn coord_binary_search<'a>(
     while size > 1 {
         let half = size / 2;
         let mid = base + half;
-        let v = coords.get(mid as usize).coord();
+        let v = coords.get(mid as usize).coord;
```
CyanRook: v -> value?

rust-src/src/gridstore/spatial.rs
```diff
@@ -174,15 +173,15 @@ fn coord_binary_search<'a>(
...
         let cmp = v.cmp(&val);
         base = if cmp == Less { base } else { mid };
```
CyanRook: Can't this be one line?

rust-src/src/gridstore/store.rs
```diff
@@ -1,138 +1,80 @@
...
+fn decode_value<T: AsRef<[u8]>>(value: T) -> impl Iterator<Item = GridEntry> {
+    let record_ref = {
+        let value_ref: &[u8] = value.as_ref();
+        // this is pretty sketch: we're opting out of compiler lifetime protection
+        // for this reference. This usage should be safe though, because we'll move the
+        // reference and the underlying owned object around together as a unit (the
+        // tuple below) so that when we pull the reference into the inner closures,
+        // we'll drag the owned object along, and won't drop it until the whole
+        // nest of closures is deleted
+        let static_ref: &'static [u8] = unsafe { std::mem::transmute(value_ref) };
```
CyanRook: We should have a ticket to revisit this at some point, but seems fine for now.

rust-src/src/gridstore/store.rs
```diff
@@ -1,138 +1,80 @@
...
+        (value, static_ref)
+    };
+    let reader = gridstore_format::Reader::new(record_ref.1);
+    let record = { gridstore_format::read_phrase_record_from(&reader) };
+
+    let iter = gridstore_format::read_var_vec_raw(record_ref.1, record.relev_scores)
+        .into_iter()
+        .flat_map(move |rs_obj| {
+            // grab a reference to the outer object to make sure it doesn't get freed
+            let _ref = &record_ref;
```
CyanRook: This feels weird.  Why would it get freed?

rust-src/src/gridstore/store.rs
```diff
@@ -1,138 +1,80 @@
...
+
+            let relev_score = rs_obj.relev_score;
+            let relev = relev_int_to_float(relev_score >> 4);
+            // mask for the least significant four bits
+            let score = relev_score & 15;
```
CyanRook: Relevance & Score should be abstracted.  We have several locations where relevance and score are getting extracted via bit manipulation and will be easy to make mistakes updating.

rust-src/src/gridstore/store.rs
```diff
@@ -144,44 +86,192 @@ fn decode_value<T: AsRef<[u8]>>(value: T) -> impl Iterator<Item = GridEntry> {
...
+                scoredist1.partial_cmp(scoredist2).unwrap() == Ordering::Greater
+            });
+
+            let nested_ref = record_ref.1;
+            all_coords.flat_map(
+                move |(distance, within_radius, score, scoredist, x, y, coords_obj)| {
+                    let ids = gridstore_format::read_fixed_vec_raw(nested_ref, coords_obj.ids);
+
+                    ids.into_iter().map(move |id_comp| {
+                        let id = id_comp >> 8;
```
CyanRook: Again, these repeated bit manipulations should be abstracted

rust-src/src/gridstore/store.rs
```diff
@@ -192,231 +282,69 @@ impl GridStore {
...
+            MatchPhrase::Exact(id) => (id, id + 1, 0),
             MatchPhrase::Range { start, end } => {
-                let remainder = start % 1024;
-                let prefix_start = if remainder == 0 { start } else { start + (1024 - remainder) };
-                let remainder = end % 1024;
-                let prefix_end = if remainder == 0 { end } else { end - remainder };
-                if prefix_start >= prefix_end {
-                    range_list.push((start, end, 0));
+                if self.bin_boundaries.contains(&start) && self.bin_boundaries.contains(&end) {
+                    (start, end, 1)
```
CyanRook: What do 0, 1 signify? Can we make constants to represent this so it is clear?

test_utils/src/lib.rs
```diff
@@ -133,13 +154,19 @@ pub fn ensure_downloaded(datafile: &str) -> PathBuf {
...
-    std::fs::create_dir_all(&tmp);
+    std::fs::create_dir_all(&tmp).unwrap();
     let idx_path = tmp.join(Path::new(&datafile.replace(".dat.lz4", ".rocksdb")));
     if !idx_path.exists() {
-        let dl_path = ensure_downloaded(datafile);
-        let decoder = Decoder::new(File::open(dl_path).unwrap()).unwrap();
-        let file = io::BufReader::new(decoder);
-        load_db_from_json_reader(file, idx_path.to_str().unwrap());
+        let grid_path = ensure_downloaded(datafile);
+        let splits_path = ensure_downloaded(&datafile.replace(".gridstore.dat.lz4", ".splits.lz4"));
```
CyanRook: What do these extensions signifiy? Can we turn these strings into named constants to make it more clear?

benches/prod_data.rs
```diff
@@ -116,7 +116,7 @@ pub fn benchmark(c: &mut Criterion) {
         records
     });
     to_bench.push(Fun::new("builder_append_logged", move |b: &mut Bencher, _i| {
-        Lazy::force(&asi_records);
+        Lazy::force(&us_records);
```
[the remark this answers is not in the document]
apendleton: I think because asiopen-place was too small to produce meaningful bench results, so we needed something bigger.

native/src/lib.rs
```diff
@@ -94,7 +94,48 @@ declare_types! {
...
+                }
+            };
+
+            match insert {
+                Ok(_) => Ok(JsUndefined::new().upcast()),
+                Err(e) => cx.throw_type_error(e)
+            }
+        }
+
+        method compactAppend(mut cx) {
```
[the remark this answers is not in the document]
apendleton: `cx` is the convention within Neon and used in all their docs for this thing. See, e.g., https://github.com/neon-bindings/examples/blob/master/classes/native/src/lib.rs . Would prefer to keep it consistent with upstream usage.

native/src/lib.rs
```diff
@@ -141,6 +182,39 @@ declare_types! {
             }
         }
 
+        method loadBinBoundaries(mut cx) {
+            let bin_boundaries = cx.argument::<JsArrayBuffer>(0)?;
+            let mut this = cx.this();
```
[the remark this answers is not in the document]
apendleton: Hm, I disagree. `this` is what this thing is called in Javascript; we only use it for that purpose here and it has no fixed meaning in Rust (where a similar concept is instead called `self`).

rust-src/src/gridstore/builder.rs
```diff
@@ -1,99 +1,152 @@
...
 /// Extends a BuildEntry with the given values.
-fn extend_entries(builder_entry: &mut BuilderEntry, values: &[GridEntry]) -> () {
-    for (rs, values) in
-        &values.into_iter().group_by(|value| (relev_float_to_int(value.relev) << 4) | value.score)
-    {
-        let rs_entry = builder_entry.entry(rs).or_insert_with(|| BTreeMap::new());
-        for (zcoord, values) in
-            &values.into_iter().group_by(|value| interleave_morton(value.x, value.y))
+fn extend_entries(builder_entry: &mut BuilderEntry, values: Vec<GridEntry>) -> () {
+    for (rs, rs_values) in somewhat_eager_groupby(values.into_iter(), |value| {
```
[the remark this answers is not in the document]
apendleton: `rs` is a composite relevance and score; `rs_values` are the set of coordinates and IDs that correspond to each relevance and score.

rust-src/src/gridstore/builder.rs
```diff
@@ -1,99 +1,152 @@
...
+        (relev_float_to_int(value.relev) << 4) | value.score
+    }) {
+        let rs_entry =
+            builder_entry.entry(rs).or_insert_with(|| HashMap::with_capacity(rs_values.len()));
+        for (zcoord, zc_values) in
```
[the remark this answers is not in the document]
apendleton: Yeah, Morton-encoded coordinates.

rust-src/src/gridstore/builder.rs
```diff
@@ -1,99 +1,152 @@
...
-    let mut fb_builder = flatbuffers::FlatBufferBuilder::new();
-    let mut rses: Vec<_> = Vec::new();
-    for (rs, coord_group) in value.iter_mut().rev() {
-        let mut coords: Vec<_> = Vec::new();
-        for (coord, ids) in coord_group.iter_mut().rev() {
+fn get_encoded_value(value: BuilderEntry) -> Result<Vec<u8>, Error> {
+    let mut builder = gridstore_format::Writer::new();
+
+    let mut items: Vec<(_, _)> = value.into_iter().collect();
+    items.sort_by(|a, b| b.0.cmp(&a.0));
```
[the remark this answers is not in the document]
apendleton: A `BuilderEntry` is a `HashMap<u8, HashMap<u32, SmallVec<[u32; 4]>>>`, so, a hashmap from a relevance/score to a hashmap of a coordinate to a vector of IDs. I'll see if I can name these differently to make it more clear.

rust-src/src/gridstore/builder.rs
```diff
@@ -1,99 +1,152 @@
...
     /// Inserts a new GridStore entry with the given values.
-    pub fn insert(&mut self, key: &GridKey, values: &[GridEntry]) -> Result<(), Error> {
+    pub fn insert(&mut self, key: &GridKey, values: Vec<GridEntry>) -> Result<(), Error> {
         let mut to_insert = BuilderEntry::new();
         extend_entries(&mut to_insert, values);
         self.data.insert(key.to_owned(), to_insert);
         Ok(())
     }
 
     ///  Appends a values to and existing GridStore entry.
```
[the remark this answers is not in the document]
apendleton: This is just a philosophical difference. Mapbox eng. standards favor docstrings for all functions. 🤷‍♂

rust-src/src/gridstore/builder.rs
apendleton: We call it `finish` because fuzzy-phrase calls its equivalent mechanism `finish`, and it in turn calls it that because `fst`, which heavily inspired its API, calls it `finish`. I'm generally pro-consistency where possible.

rust-src/src/gridstore/store.rs
apendleton: This is the price we paid for the static-cast hackery. We told it our references will live forever (cast them to `'&'static`) but they won't, they'll only live as long as `record_ref`. We're going to return this whole iterator from this function, so the closure will outlive the outer function's stack, which is where `record_ref` will live unless we pull it into the closure -- this is a `move` closure, so referring to record_ref here, even though we're not doing anything with it, moves it in and ensures that it continues to exist for the life of the iterator.

test_utils/src/lib.rs
apendleton: The first is a file full of grids, and the second is a file full of prefix boundaries. We can make them constants, though, sure.

native/src/lib.rs
```diff
@@ -141,6 +182,39 @@ declare_types! {
...
+                            }
+                            None => {
+                                Err("can't call loadBinBoundaries after finish()".to_owned())
+                            }
+                        }
+                    },
+                    Err(e) => Err(e.to_string())
+                };
+
+                borrow_result
```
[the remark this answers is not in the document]
apendleton: You'd think, but just returning the match doesn't compile:

1__andrew_taverner___Users_andrew_Work_carmen-core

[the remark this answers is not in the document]
CyanRook: Ah okay, at least the error message is helpful. It seems like the borrow checker can't make the assertion that `&lock` will be freed until after the match drops it which is why it is requiring the explicit dropping of the borrow before returning which is why it becomes it's own variable whose only purpose is to let the match drop.

rust-src/src/gridstore/common.rs
apendleton: Hm, there's a mismatched paren there, so I'm not sure I follow, but I think the last paren is extraneous and what you're intending is `(orig_bbox[2] + 1) * scale_multiplier - 1`. If that's the case, this comment explains why we can't do that: `(orig_bbox[2] + 1) * scale_multiplier` might overflow. Assuming we start with the easternomst z15 X coordinate 32767 and we're scaling to z16, our scale_multiplier will be 2, and the output we want is the rightmost z16 coordinate, 65535 -- this is also the maximum allowed unsigned 16-bit int. As currently formulated, we calculate it as:
```
32767 * 2 + (2 - 1)
= 65534 + 1
= 65535
```
but what you've proposed is
```
(32767 + 1) * 2 - 1
= 32768 * 2 - 1
= 65536 - 1
= 65535
```
... but `65536` doesn't fit in a u16.

apendleton: Hm, there's a mismatched paren there, so I'm not sure I follow, but I think the last paren is extraneous and what you're intending is `(orig_bbox[2] + 1) * scale_multiplier - 1`. If that's the case, this comment explains why we can't do that: `(orig_bbox[2] + 1) * scale_multiplier` might overflow. Assuming we start with the easternomst z15 X coordinate 32767 and we're scaling to z16, our scale_multiplier will be 2, and the output we want is the rightmost z16 coordinate, 65535 -- this is also the maximum allowed unsigned 16-bit int. As currently formulated, we calculate it as:
```
32767 * 2 + (2 - 1)
= 65534 + 1
= 65535
```
but what you've proposed is
```
(32767 + 1) * 2 - 1
= 32768 * 2 - 1
= 65536 - 1
= 65535
```
... but `65536` doesn't fit in a u16.

rust-src/src/gridstore/builder.rs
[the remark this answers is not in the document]
apendleton: I've now renamed all the things being compared

rust-src/src/gridstore/builder.rs
```diff
@@ -1,99 +1,152 @@
...
-    for (rs, coord_group) in value.iter_mut().rev() {
-        let mut coords: Vec<_> = Vec::new();
-        for (coord, ids) in coord_group.iter_mut().rev() {
+fn get_encoded_value(value: BuilderEntry) -> Result<Vec<u8>, Error> {
+    let mut builder = gridstore_format::Writer::new();
+
+    let mut items: Vec<(_, _)> = value.into_iter().collect();
+    items.sort_by(|a, b| b.0.cmp(&a.0));
+
+    let mut rses: Vec<_> = Vec::with_capacity(items.len());
```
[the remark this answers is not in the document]
apendleton: done

rust-src/src/gridstore/builder.rs
```diff
@@ -1,99 +1,152 @@
...
+
+    let mut id_lists: HashMap<_, gridstore_format::FixedVecOffset<u32>> = HashMap::new();
+
+    for (rs, coord_group) in items.into_iter() {
+        let mut inner_items: Vec<(_, _)> = coord_group.into_iter().collect();
+        inner_items.sort_by(|a, b| b.0.cmp(&a.0));
```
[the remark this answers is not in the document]
apendleton: I've now renamed all the things being compared

rust-src/src/gridstore/builder.rs
```diff
@@ -1,99 +1,152 @@
...
+
+        let mut coords: Vec<_> = Vec::with_capacity(inner_items.len());
+
+        for (coord, mut ids) in inner_items.into_iter() {
             // reverse sort
             ids.sort_by(|a, b| b.cmp(a));
```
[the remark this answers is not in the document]
apendleton: I've now renamed all the things being compared

rust-src/src/gridstore/coalesce.rs
apendleton: Moved to a constant in common.rs

rust-src/src/gridstore/spatial.rs
apendleton: Renamed these to be clearer

rust-src/src/gridstore/spatial.rs
apendleton: Renamed

rust-src/src/gridstore/spatial.rs
apendleton: Yeah, honestly I'm not sure why .cmp() is being explicitly called here -- my only change in this branch was to drop the parens on coord as it's no longer a function, but looking now, this not very idiomatic. I went through and got rid of the explicit cmp calls here and elsewhere, in favor of just using <, >, and ==, and generally tightened things up a bit.

rust-src/src/gridstore/store.rs
apendleton: These are type markers; we use the first byte in each rocksdb key to signal if it's a single phrase or a bin. I moved these to an enum with explicit discriminators, here and elsewhere.

test_utils/src/lib.rs
apendleton: done

rust-src/src/gridstore/store.rs
apendleton: ticketed in #61

rust-src/src/gridstore/store.rs
apendleton: ticketed in #61

rust-src/src/gridstore/builder.rs
```diff
@@ -1,99 +1,152 @@
...
+        relev: f64,
+        score: u8,
+        id: u32,
+        source_phrase_hash: u8,
+        coords: &[(u16, u16)],
+    ) {
+        let to_append =
+            self.data.entry(key.to_owned()).or_insert_with(|| BuilderEntry::with_capacity(1));
+
+        let relev_score = (relev_float_to_int(relev) << 4) | score;
```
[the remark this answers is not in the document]
apendleton: ticketed in #61

rust-src/src/gridstore/gridstore_format.rs
apendleton: ticketed in #61

rust-src/src/gridstore/common.rs
CyanRook: In production, rust will wrap the value, but it might panic on tests.  Either way your explanation makes sense and no need to change it.

rust-src/src/gridstore/store.rs
apendleton: ticketed in #62

rust-src/src/gridstore/builder.rs
CyanRook: Seems like an odd convention, but since it's already in place in fst & fuzzy phrase, it seems fine.  Interestingly, looking into fst, all it does is flush the writer by calling .into_iter().  The writing to disk comment in the docs is only there because the writer happens to be writing to disk.

native/src/lib.rs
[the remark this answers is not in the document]
CyanRook: I'd still prefer js_context_this especially since we are doing this whole cross language code stuff and `this` is such a common construct across languages it seems easy to be confused in the future, but since its hidden in the neon bindings, I don't care enough to push it.

native/src/lib.rs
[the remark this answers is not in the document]
CyanRook: Ah okay, if we are aligning to crate docs then that is good enough.

PR #67, opened 2020-02-25
clean up

aarthykc:
Per next actions here: #64 
1. This PR cleans up the `PhrasematchResults` struct, removing parameters that we won't be using at all 
2. Cleans up a few of the variable names in functions (there could be more there)
3. Cleans up the test for making sure deserialize_js_phrasematchresults and js_stackable work as expected
4. Changes where nmask, idx and bmask get deserialised (since it happens for phrasematch results per index)

cc @CyanRook

### DIFF
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -1,4 +1,5 @@
 use carmen_core::gridstore::coalesce;
+use carmen_core::gridstore::stackable;
 use carmen_core::gridstore::PhrasematchSubquery;
 use carmen_core::gridstore::{
     CoalesceContext, GridEntry, GridKey, GridStore, GridStoreBuilder, MatchOpts, MatchKey, PhrasematchResults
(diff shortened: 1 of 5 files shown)


native/src/lib.rs
```diff
@@ -441,86 +442,72 @@ where
...
-                phrase: neon_serde::from_value(cx, phrase)?,
                 scorefactor: neon_serde::from_value(cx, scorefactor)?,
                 prefix: neon_serde::from_value(cx, prefix)?,
                 weight: neon_serde::from_value(cx, weight)?,
                 match_key: MatchKey { match_phrase: neon_serde::from_value(cx, match_phrase)?, lang_set },
                 idx: neon_serde::from_value(cx, idx)?,
                 zoom: neon_serde::from_value(cx, zoom)?,
                 nmask: neon_serde::from_value(cx, nmask)?,
-                nmask: neon_serde::from_value(cx, mask)?,
+                mask: neon_serde::from_value(cx, mask)?,
```
CyanRook: How hard would it be in Rust only to change these names to something other than mask, nmask, and bmask to be something more meaningful and useful?  We don't need to change the JS side (yet), just rust.

rust-src/src/gridstore/stackable.rs
```diff
@@ -28,10 +33,16 @@ pub fn stackable<'a, T: Borrow<GridStore> + Clone + Debug>(phrasematch_results:
...
-            let target_mask = &phrasematches.idx | node.mask;
-            node.children.push(stackable(&phrasematch_results, Some(phrasematches.clone()), target_nmask, phrasematches.clone().bmask, target_mask));
+            if (node.nmask & phrasematches.nmask) == 0 && (node.mask & phrasematches.mask) == 0 {
+                let target_nmask = &phrasematches.nmask | node.nmask;
+                let target_mask = &phrasematches.idx | node.mask;
+                node.children.push(stackable(
+                    &phrasematch_results,
+                    Some(phrasematches.clone()),
+                    target_nmask,
+                    phrasematches.clone().bmask,
```
CyanRook: Is the bmask for the next one just the bmask of that node or some modification based on the history of bmask (parents bmask)?

rust-src/src/gridstore/stackable.rs
```diff
@@ -28,10 +33,16 @@ pub fn stackable<'a, T: Borrow<GridStore> + Clone + Debug>(phrasematch_results:
...
-            if (node.nmask & phrasematches.nmask) == 0  && (node.mask & phrasematches.mask) == 0 {
-            let target_nmask = &phrasematches.nmask | node.nmask;
-            let target_mask = &phrasematches.idx | node.mask;
-            node.children.push(stackable(&phrasematch_results, Some(phrasematches.clone()), target_nmask, phrasematches.clone().bmask, target_mask));
+            if (node.nmask & phrasematches.nmask) == 0 && (node.mask & phrasematches.mask) == 0 {
+                let target_nmask = &phrasematches.nmask | node.nmask;
+                let target_mask = &phrasematches.idx | node.mask;
+                node.children.push(stackable(
+                    &phrasematch_results,
+                    Some(phrasematches.clone()),
```
CyanRook: Is there a way to get ride of cloning or not really? How slow does it make it?  There might be a way using Rc::<> over the individual phrase matches so that your type becomes Vec<Vec<Rc<PhraseMatch>>>.

rust-src/src/gridstore/stackable.rs
aarthykc: nope, the modification is definitely something I need to do, I've added it as a part of next actions

native/src/lib.rs
aarthykc: It shouldn't be hard to change it, y, nmask and bmask are not helpful at all

rust-src/src/gridstore/stackable.rs
aarthykc: Added in next actions here - https://github.com/mapbox/carmen-core/pull/64

rust-src/src/gridstore/stackable.rs
aarthykc: Added  next action here - https://github.com/mapbox/carmen-core/pull/64

native/src/lib.rs
aarthykc: Added in next actions here - https://github.com/mapbox/carmen-core/pull/64

PR #68, opened 2020-03-02
Use bmask for evaluation and calculate max relevance for root node

aarthykc:
Per next action #64 - 

> Add bmask in the evaluation of whether stacks can align

Additionally, we also need the max relevance which is the sum of the node's relevance and the maximum of the children's relevances added to the parent node in order to be able to make intelligent decisions while traversing the tree. This PR calculates and adds the max_relevance property to the parent node and uses the bmask to evaluate whether stacks can align.

We also need a way to prevent duplicate branches by allowing stacking only in one direction otherwise, we could end up with a tree that has phrasematch results stacked as country -> postcode as well as postcode -> country. This PR also adds a condition to allow stacking to only happen only in one direction to prevent duplicate branches.

### DIFF
--- a/rust-src/src/gridstore/stackable.rs
+++ b/rust-src/src/gridstore/stackable.rs
@@ -49,10 +49,11 @@ pub fn stackable<'a, T: Borrow<GridStore> + Clone + Debug>(
     for phrasematch_per_index in phrasematch_results.iter() {
         for phrasematches in phrasematch_per_index.iter() {
             if node.zoom > phrasematches.zoom {
-                if phrasematches.idx >= node.idx {
+                continue;
+            } else if node.zoom == phrasematches.zoom {
+                if node.idx < phrasematches.idx {
                     continue;
                 }
-                continue;
             }
 
             if (node.nmask & phrasematches.nmask) == 0
--- a/rust-src/src/gridstore/stackable.rs
+++ b/rust-src/src/gridstore/stackable.rs
@@ -20,6 +20,7 @@ pub struct StackableNode<T: Borrow<GridStore> + Clone + Debug> {
     pub idx: u32,
     pub max_relev: f64,
     pub adjusted_relev: f64,
+    pub zoom: u16,
 }
 
 pub fn stackable<'a, T: Borrow<GridStore> + Clone + Debug>(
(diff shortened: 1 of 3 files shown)


rust-src/src/gridstore/stackable.rs
```diff
@@ -22,30 +27,56 @@ pub fn stackable<'a, T: Borrow<GridStore> + Clone + Debug>(
...
+                continue;
+            }
+            if (node.nmask & phrasematches.nmask) == 0
+                && (node.mask & phrasematches.mask) == 0
+                && phrasematches.bmask.contains(&node.idx) == false
+            {
                 let target_nmask = &phrasematches.nmask | node.nmask;
-                let target_mask = &phrasematches.idx | node.mask;
+                let target_mask = &phrasematches.mask | node.mask;
+                let target_bmask = &phrasematches.bmask;
```
CyanRook: bmask needs to be updated based on the previous nodes bmask and the to be added phrase matches bmask.

rust-src/src/gridstore/stackable.rs
```diff
@@ -22,30 +27,56 @@ pub fn stackable<'a, T: Borrow<GridStore> + Clone + Debug>(
...
 
     for phrasematch_per_index in phrasematch_results.iter() {
         for phrasematches in phrasematch_per_index.iter() {
-            if (node.nmask & phrasematches.nmask) == 0 && (node.mask & phrasematches.mask) == 0 {
+            if phrasematches.idx >= node.idx {
+                continue;
+            }
+            if (node.nmask & phrasematches.nmask) == 0
+                && (node.mask & phrasematches.mask) == 0
+                && phrasematches.bmask.contains(&node.idx) == false
```
CyanRook: Since bmask is only being used for lookup, a (hash)set would be more appropriate than a vector

rust-src/src/gridstore/stackable.rs
```diff
@@ -40,20 +43,27 @@ pub fn stackable<'a, T: Borrow<GridStore> + Clone + Debug>(
...
         max_relev: max_relev,
         adjusted_relev: adjusted_relev,
+        zoom: zoom,
     };
 
     for phrasematch_per_index in phrasematch_results.iter() {
         for phrasematches in phrasematch_per_index.iter() {
-            if phrasematches.idx >= node.idx {
+            if node.zoom > phrasematches.zoom {
+                if phrasematches.idx >= node.idx {
```
CyanRook: This logic is not correct.  As written, the phrasematches.idx portion is irrelevant due to the continue after it.

```
if node.zoom > phrasematches.zoom {
    continue;
} else if node.zoom == phrasematches.zoom {
    if node.idx < phrasematches.idx {
        continue;
    }
}
```

The above could be simplified into a single expression, but doesn't really matter.

PR #69, opened 2020-03-11
Port unit tests from stackable javascript over to carmen-core to ensure feature parity

aarthykc:
## Context
As a part of modifying stackable to be able to allow comparison of more than 40 candidates (which is our current cut off), this PR ports the unit tests over from stackable in javascript over to rust to ensure feature parity between the old and new version of stackable. This PR also adds a basic breath first search tree traversal which is to be used only in tests to be able to make assertions about the results at the end.

cc @apendleton @CyanRook

### DIFF
--- a/rust-src/src/gridstore/stackable.rs
+++ b/rust-src/src/gridstore/stackable.rs
@@ -193,11 +193,26 @@ mod test {
         let phrasematch_results = vec![vec![a1, b1, b2]];
 
         let tree = stackable(&phrasematch_results, None, 0, HashSet::new(), 0, 129, 0.0, 0.0, 0, 0);
-        let a1_children_ids: Vec<u32> = tree.clone().children[0].clone().children.iter().map(|node| node.phrasematch_id).collect();
+        let a1_children_ids: Vec<u32> = tree.clone().children[0]
+            .clone()
+            .children
+            .iter()
+            .map(|node| node.phrasematch_id)
+            .collect();
         assert_eq!(vec![1, 2], a1_children_ids, "a1 can stack with b1 and b2");
-        let b1_children_ids: Vec<u32> = tree.clone().children[1].clone().children.iter().map(|node| node.phrasematch_id).collect();
+        let b1_children_ids: Vec<u32> = tree.clone().children[1]
+            .clone()
+            .children
+            .iter()
+            .map(|node| node.phrasematch_id)
+            .collect();
         assert_eq!(0, b1_children_ids.len(), "b1 cannot stack with b2, same nmask");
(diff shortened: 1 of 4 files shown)


rust-src/src/gridstore/stackable.rs
```diff
@@ -158,6 +184,129 @@ mod test {
...
+        assert_eq!(
+            result_ids,
+            [129, 1, 1, 0, 1, 1],
+            "idx stacked in the right order, no duplicate branches"
+        );
+        let tree = stackable(&phrasematch_results, None, 0, HashSet::new(), 0, 129, 0.0, 0.0, 0);
+        let result_max_relev: Vec<f64> = bfs(tree).iter().map(|node| node.max_relev).collect();
+        assert_eq!(
+            result_max_relev,
+            [1.0, 0.5, 0.5, 1.0, 0.5, 0.5],
```
CyanRook: It seems like we should use some different values for relevance to help differentiate what is going on, unless each node has a unique id.

rust-src/src/gridstore/stackable.rs
```diff
@@ -158,6 +184,129 @@ mod test {
         };
 
         let phrasematch_results = vec![vec![a1, b1, b2]];
-        stackable(&phrasematch_results, None, 0, HashSet::new(), 0, 129, 0.0, 0.0, 0);
+        let tree = stackable(&phrasematch_results, None, 0, HashSet::new(), 0, 129, 0.0, 0.0, 0);
+        let result_ids: Vec<u32> = bfs(tree).iter().map(|node| node.idx).collect();
+        assert_eq!(
+            result_ids,
+            [129, 1, 1, 0, 1, 1],
```
CyanRook: These are the indexes, not a node id, correct?  Maybe it would be useful to add a 
```rust
#[cfg(test)]
node_id: usize
```
To the phrasematch to uniquely identify them during testing so you can see what actually happened, instead of relying on nonunique characteristics.

rust-src/src/gridstore/stackable.rs
```diff
@@ -96,8 +121,9 @@ pub fn stackable<'a, T: Borrow<GridStore> + Clone + Debug>(
 #[cfg(test)]
 mod test {
     use super::*;
+
     #[test]
-    fn stackable_test() {
+    fn simple_stackable_test() {
```
CyanRook: This test is hard to follow and it's unclear if the result is what is expected.  Adding some test only debug information via `#[cfg(test)]` or `#[cfg(debug)]` to help make it clear could be helpful to ensure the right items end up in the right place.

PR #70, opened 2020-03-11
Stack and coalesce

apendleton:
Two big things in here:

## A new version of coalesce that operates on the trees produced by the new stackable function

There's a new function, `tree_coalesce` that operates on stacks. It identifies what would have been single-item stacks (basically, descendants of the root that have no children), and does something similar to the old coalesce_single, which does a very limited index scan, and for all other cases, instead of the old coalesce_multi, it does a recursive exploration of the descendants of the root, accumulating state at each recursion step.

At present this function isn't parallel at all, and exhaustively explores the entire tree, which we'll want to change in subsequent PRs.

## A combined stack-and-coalesce function

In Carmen we want to move to having spatialmatch call a function with all of its phrasematches that does both the stackable work and the coalesce work, so this branch adds that composite function, `stackable_and_coalesce` (camel case in the binding). Right now it's pretty basic, but eventually we'll want to do some deduping, etc. It also adds bindings for JS.

In addition to all that, there are some extra odds and ends to match new expectations on the Carmen side or to get Carmen tests to pass.

### DIFF
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -9,7 +9,7 @@ use itertools::Itertools;
 use ordered_float::OrderedFloat;
 
 use crate::gridstore::common::*;
-use crate::gridstore::stackable::StackableNode;
+use crate::gridstore::stackable::{stackable, StackableNode};
 use crate::gridstore::store::GridStore;
 
 /// Takes a vector of phrasematch subqueries (stack) and match options, gets matching grids, sorts the grids,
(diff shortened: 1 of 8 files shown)


rust-src/src/gridstore/coalesce.rs
```diff
@@ -325,3 +328,269 @@ fn coalesce_multi<T: Borrow<GridStore> + Clone>(
...
+                    .entry((grid.grid_entry.x, grid.grid_entry.y))
+                    .or_insert_with(|| vec![]);
+                state_vec.push(context);
+            }
+
+            let mut multi_contexts = Vec::new();
+            tree_recurse(&node, match_opts, &prev_state, subquery.zoom, &mut multi_contexts)?;
+
+            // penalize singnle-entry stacks and ascending stacks for... some reason?
+            for mut context in multi_contexts {
```
aarthykc: @apendleton not a blocker for now: but we should probably dig up the reasoning behind having this penalty in the first place?

aarthykc:
@apendleton I can make sure tests pass in a following PR.

rust-src/src/gridstore/coalesce.rs
apendleton: @aarthykc I actually think we should probably drop the penalty, but I added it at least temporarily because I had tests that made sure that at least in most cases, old-coalesce and new-coalesce agree. Once old-coalesce gets deleted that will be less of a motivation.

PR #72, opened 2020-03-16
Refactor phrasematch types

apendleton:
This branch does a few of main things:
- [x] recombines the PhrasematchSubquery and PhrasematchResults types back into one type that everything can use
- [x] moves all the properties that are always the same for every phrasematch for a given store into the store and out of the phrasematch type (so, idx, zoom, etc.)
- [x] flatten the input for stackable and stack_and_coalesce to just be an array of phrasematches, rather than an array of arrays

One other thing: it looks like since I had forked off, a parameter was added to stackable stackable and a field was added to StackableNode that had the phrasematch ID in it. Per chat, I'm using the phrasematch ID field on the phrasematch instead here.

After that, most of what's here is just propagating those changes outwards. Some comments inline below.

### DIFF
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -462,8 +462,6 @@ where
             gridstore_clone
         };
         let weight = js_phrasematch.get(cx, "weight")?;
-        let idx = js_phrasematch.get(cx, "idx")?;
-        let zoom = js_phrasematch.get(cx, "zoom")?;
         let mask = js_phrasematch.get(cx, "mask")?;
 
         let match_key = js_phrasematch.get(cx, "match_key")?.downcast::<JsObject>().or_throw(cx)?;
@@ -504,9 +502,6 @@ fn deserialize_phrasematch_results<'j, C: Context<'j>>(
     for i in 0..js_phrasematch_per_index.len() {
         let js_phrasematch = js_phrasematch_per_index.get(cx, i)?.downcast::<JsObject>().or_throw(cx)?;
         let phrasematch_array = js_phrasematch.get(cx, "phrasematches")?.downcast::<JsArray>().or_throw(cx)?;
-        let nmask = js_phrasematch.get(cx, "nmask")?;
-        let idx = js_phrasematch.get(cx, "idx")?;
-        let bmask = js_phrasematch.get(cx, "bmask")?;
 
         let phrasematch_array_length = phrasematch_array.len();
         let mut phrasematches: Vec<PhrasematchSubquery<ArcGridStore>> = Vec::with_capacity(phrasematch_array_length as usize);
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -1,7 +1,6 @@
 use carmen_core::gridstore::{coalesce, stackable, stack_and_coalesce};
-use carmen_core::gridstore::PhrasematchSubquery;
 use carmen_core::gridstore::{
-    CoalesceContext, GridEntry, GridKey, GridStore, GridStoreBuilder, MatchOpts, MatchKey, PhrasematchResults
+    CoalesceContext, GridEntry, GridKey, GridStore, GridStoreBuilder, MatchOpts, MatchKey, PhrasematchSubquery
 };
 
 use neon::prelude::*;
@@ -46,7 +45,7 @@ impl Task for CoalesceTask {
 }
 
 struct StackAndCoalesceTask {
-    argument: (Vec<Vec<PhrasematchResults<ArcGridStore>>>, MatchOpts),
+    argument: (Vec<Vec<PhrasematchSubquery<ArcGridStore>>>, MatchOpts),
 }
 
 impl Task for StackAndCoalesceTask {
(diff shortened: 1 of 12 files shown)


native/src/lib.rs
```diff
@@ -488,76 +511,13 @@ where
...
-                subquery_edit_distance: neon_serde::from_value(cx, subquery_edit_distance)?,
-                id: neon_serde::from_value(cx, id)?,
-            };
-            phrasematches.push(phrasematch_result);
-        }
-        phrasematch_results_by_index.push(phrasematches);
-    }
-    Ok(phrasematch_results_by_index)
-}
-
```
apendleton: This function is now duplicative of the other deserialize function, since they both just unpack vectors of phrasematches.

rust-src/src/gridstore/common.rs
```diff
@@ -94,16 +93,10 @@ impl MatchKey {
...
-pub struct Proximity {
-    pub point: [u16; 2],
-    pub radius: f64,
-}
-
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub struct MatchOpts {
     pub bbox: Option<[u16; 4]>,
-    pub proximity: Option<Proximity>,
+    pub proximity: Option<[u16; 2]>,
```
apendleton: radius is now a property on the whole store, so this can just be a two-member array rather than a complex type with both a point and a radius

rust-src/src/gridstore/store.rs
```diff
@@ -20,6 +20,12 @@ pub struct GridStore {
     db: DB,
     bin_boundaries: HashSet<u32>,
     pub path: PathBuf,
+    // options:
+    pub idx: u16,
+    pub zoom: u16,
+    pub type_id: u16,
+    pub non_overlapping_indexes: HashSet<u16>, // the field formerly known as bmask
```
apendleton: I tried to give this a more descriptive name than it had before.

rust-src/src/gridstore/stackable.rs
```diff
@@ -140,78 +129,57 @@ mod test {
         ];
         builder.insert(&key, entries).expect("Unable to insert record");
         builder.finish().unwrap();
-        let store = GridStore::new(directory.path()).unwrap();
+        let store1 =
+            GridStore::new_with_options(directory.path(), 1, 14, 1, HashSet::new(), 200.).unwrap();
+        let store2 =
+            GridStore::new_with_options(directory.path(), 2, 14, 2, HashSet::new(), 200.).unwrap();
```
apendleton: Now all phrasematches that are from the same index by necessity have the same idx and the same type ID (which becomes the nmask), so in order to actually get these to stack I had to have two different stores.

aarthykc: I don't think that I completely get this comment, we create a new store per index even now right? Are you saying for the sake of tests and to be able to demonstrate being able to stack (since they have the bmask and ndx information now) we'll have to create two stores?

rust-src/src/gridstore/store.rs
```diff
@@ -20,6 +20,12 @@ pub struct GridStore {
     db: DB,
     bin_boundaries: HashSet<u32>,
     pub path: PathBuf,
+    // options:
+    pub idx: u16,
+    pub zoom: u16,
+    pub type_id: u16,
```
aarthykc: and type_id is the nmask?

rust-src/src/gridstore/store.rs
aarthykc: y, reads much better!

rust-src/src/gridstore/store.rs
```diff
@@ -245,6 +251,17 @@ impl<T: Iterator<Item = MatchEntry>> Eq for QueueElement<T> {}
 
 impl GridStore {
     pub fn new<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
+        GridStore::new_with_options(path, 0, 6, 0, HashSet::new(), 0.0)
```
aarthykc: Won't they all have options? should `new` be replaced by `new_with_options`?

rust-src/src/gridstore/stackable.rs
apendleton: @aarthykc basically, the way the test was before, it was trying to get successful stacking from two different phrasematches that both used the same store. That worked before, because even though the `store` was the same across the different phrasematches, the `idx` and `nmask` were different. Now that `idx` and the equivalent of `nmask` have moved out of the phrasematch and into the store, if the store is the same, the idx and nmask will also be the same, and stackable won't let them actually stack. That's correct behavior, but broke means this particular test has to be a little more complicated to be more like real-world usage.

rust-src/src/gridstore/store.rs
apendleton: Ish. type_id is more properly what's called `ndx` in Carmen (terrible name :cry:), so like, the first type is 0, the second type is 1, the third type is 2, etc. `nmask` is that turned into a bitmask, so rather than 0, 1, 2, 3, etc., it's 1, 2, 4, 8, etc., so that you can do bit math with them. This branch converts into that flavor on the Rust side for now, inside stackable, but if you're still planning on grouping by type in stackable you can get rid of all the nmask stuff anyway, so it seemed like it made more sense not to bake that particular concept into the interface.

rust-src/src/gridstore/store.rs
apendleton: Yeah probably... I was feeling lazy and not wanting to have to change every single constructor in the tests, so I left a `new` that sets some defaults, and figured I could just use this one in cases where the zoom or idx mattered. Turns out that was a lot, so I probably ought to go back and change the rest of them and then just have one version, but I think I might save that for a follow-up PR.

PR #74, opened 2020-03-26
Add field that indicates depth of node in the tree

aarthykc:
### Context
While coalescing the stacks it would be good to know if a node has no parent or child making it the perfect candidate for the `coalecse_single()` (coalesce's a single stack) operation. This PR adds a depth field to the node which can be used along with the function `is_leaf()`to determine whether the node is a good candidate for `coalesce _single()`. Note: The depth of the root node will always be 0. 

cc @apendleton


aarthykc:
Not needed in it's current state.

PR #75, opened 2020-03-26
[WIP] add stackable benchmarking

aarthykc:
## Context
In order to be able to ship the new version of stackable and coalesce, we're going to need performance benchmarks that we will use to profile which parts of stackable/coalesce are slow. The work here is particularly important, since in its current state - the new version of stackable doesn't have any limits which means that stackable will create a tree with all the stacks that it gets from phrasematch. We'll end up seeing huge performance regressions especially when it comes to address queries which will have a number of candidates that stackable will attempt to create a tree out of.

We'll use the benchmarks in this PR to make a call on what limits/adjustments need to be made in order to get stackable into production. 
(Will add use cases that need to be benched as I add more benchmarking data)

cc @CyanRook @apendleton


aarthykc:
A bunch has changed since this PR, closing in favour of a new PR that contains the new grid format changes and removal of coalesce benchmarks + uses collapsed_phrasematches!

PR #77, opened 2020-04-10
Stackable grouped phrasematches

aarthykc:
## Context

In order to reduce the memory footprint, we're tightening up the size of the stackable tree by grouping together phrasematches with the same idx, weight and mask. This branch attempts to group things together on the stackable end which is then consumed by coalesce. It also removes some of the unnecessary cloning in stackable by borrowing where ever possible. It also changes a couple of the tests in order to incorporate the new `MatchKeyWithId` struct, which should go away when I pull in the new changes @apendleton pushed up on the coalesce side.

cc @CyanRook @apendleton

### DIFF
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -624,7 +624,7 @@ pub fn collapse_phrasematches<T: Borrow<GridStore> + Clone + Debug>(
                     match_keys: phrasematch.match_keys,
                 };
                 entry.insert(pm);
-            },
+            }
             Entry::Occupied(mut grouped_phrasematch) => {
                 grouped_phrasematch.get_mut().match_keys.push(phrasematch.match_keys[0].clone());
             }
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -605,7 +605,7 @@ fn tree_recurse<T: Borrow<GridStore> + Clone + Debug>(
 }
 
 pub fn collapse_phrasematches<T: Borrow<GridStore> + Clone + Debug>(
-    phrasematches: &Vec<PhrasematchSubquery<T>>,
+    phrasematches: Vec<PhrasematchSubquery<T>>,
 ) -> Vec<PhrasematchSubquery<T>> {
     let mut phrasematch_results: Vec<PhrasematchSubquery<T>> = Vec::new();
     let mut phrasematch_map = HashMap::new();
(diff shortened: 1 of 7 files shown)


rust-src/src/gridstore/coalesce.rs
```diff
@@ -598,12 +604,102 @@ fn tree_recurse<T: Borrow<GridStore> + Clone + Debug>(
...
+pub fn collapse_phrasematches<T: Borrow<GridStore> + Clone + Debug>(
+    phrasematches: Vec<PhrasematchSubquery<T>>,
+) -> Vec<PhrasematchSubquery<T>> {
+    let mut phrasematch_results: Vec<PhrasematchSubquery<T>> = Vec::new();
+    let mut phrasematch_map = HashMap::new();
+    let mut group_hash;
+    for phrasematch in phrasematches.into_iter() {
+        group_hash = phrasematch.weight.to_string()
+            + &phrasematch.idx.to_string()
+            + &phrasematch.mask.to_string();
```
apendleton, line 616: Let's not use a string as a key here. It's necessary in JS but we can just use these values directly in Rust, as a tuple:
```
group_hash = (phrasematch.weight, phrasematch.idx, phrasematch.mask);
```
and it should be more performant and less risky in terms of weird collisions, etc.

rust-src/src/gridstore/coalesce.rs
```diff
@@ -598,12 +604,102 @@ fn tree_recurse<T: Borrow<GridStore> + Clone + Debug>(
...
+                non_overlapping_indexes: phrasematch.non_overlapping_indexes,
+                weight: phrasematch.weight,
+                mask: phrasematch.mask,
+                match_keys: phrasematch.match_keys,
+            };
+            phrasematch_map.insert(group_hash, pm);
+        } else {
+            let grouped_phrasematch = matched_phrasematch_group.unwrap();
+            grouped_phrasematch.match_keys.push(phrasematch.match_keys[0].clone());
+        }
```
apendleton, line 631: The business logic here looks good to me. One performance-optimization thing: the way this is constructed now, you end up hashing the same key twice, once to see if the thing is in there, and again to insert it. You can use the [Entry API](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) to avoid that (and also avoid the unwrap on 629). There are some examples in `coalesce.rs`, but essentially it ends up looking like this:
```
match some_map.entry(key) {
    Entry::Vacant(entry) => {
        let new_thing = ... // set up my new thing
        entry.insert(new_thing);
    },
    Entry::Occupied(mut already_in_there) => {
        already_in_there.some_field.push(whatever);
    }
}
```

aarthykc: Done!

rust-src/src/gridstore/coalesce.rs
aarthykc: done!

PR #78, opened 2020-04-10
[wip] Constrained coalesce

apendleton:
This PR will get cleaner once #76 lands, but the main new things are:
- [x] guts a bunch of the old benchmarks that no longer because of data format changes and s3 rearrangements
- [x] adds new benchmarks for coalesce
- [x] reworks `tree_coalesce` to be much faster:
    - [x] avoids recursion and instead maintains a priority queue of next things to work on, and stops when we've produced enough results and can't produce any more that are better than our current worst ones
    - [x] caches fetched grids so that if the same lookup occurs in multiple places in the tree, we can avoid doing the `get_matching` call again.

Still to do: tighten up the code a bit more, as it's still a bit verbose and duplicative, and maybe parallelize? Or maybe that'll wait for the next round, TBD.


apendleton:
Superseded by #83

PR #82, opened 2020-04-14
[WIP] Group phrasematches by type_id to avoid doing nmask checks

aarthykc:
## Context 

We're looking at potential speed improvements that will make stackable faster especially where us-address queries are involved. This PR attempts to bin the phrasematches by the `type_id` to avoid doing the extra check for `nmasks`. 

## Next steps
- [ ] Run benches to see if there are speed improvements
- [ ] Publish a binary to see if this works well with carmen
- [ ] Review 
- [ ] Merge 

cc @apendleton @CyanRook

### DIFF
--- a/benches/prod_data.rs
+++ b/benches/prod_data.rs
@@ -56,14 +56,16 @@ pub fn benchmark(c: &mut Criterion) {
                 let queries = prepare_stackable_phrasematches(file);
                 let collapsed: Vec<_> =
                     queries.iter().map(|q| collapse_phrasematches(q.to_vec())).collect();
-                let mut cycle = collapsed.iter().cycle();
+                let binned_phrasematches: Vec<_> =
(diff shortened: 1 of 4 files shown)


apendleton:
@aarthykc what I was going for in suggesting grouping by type was to have a different algorithm structure that recurses once per type rather than once per phrasematch. This is what makes it possible to avoid doing type/nmask checks: when you recurse, you just start with the next type, so you know that all the entries you encounter won't type-overlap with your stack so far.

It doesn't look like this branch takes that approach, from what I can tell. It still recurses once per phrasematch, and it groups by type but it doesn't look like it uses the type information, so the nmask checks are gone, but I'm not sure what functionally replaces them here (like, how you make sure you don't end up with more than one stack component of the same type in a given stack).

Let me know if that makes sense. If not, I can send you the Carmen branch from a year ago that uses this approach, or we can maybe pair on it?

PR #83, opened 2020-04-21
[wip] Parallel constrained coalesce

apendleton:
This PR supersedes #78, which tries to make coalesce faster. To review, that branch:
- [x] guts a bunch of the old benchmarks that no longer because of data format changes and s3 rearrangements
- [x] adds new benchmarks for coalesce
- [x] reworks `tree_coalesce` to be much faster:
    - [x] avoids recursion and instead maintains a priority queue of next things to work on, and stops when we've produced enough results and can't produce any more that are better than our current worst ones
    - [x] caches fetched grids so that if the same lookup occurs in multiple places in the tree, we can avoid doing the `get_matching` call again

This branch additionally:
- [x] splits the task of processing a new coalesce node into two phases, one to fetch data and one to do the aligning
- [x] grabs eight "coalesce steps" at a time from the priority queue introduced in the last round, and processes them in parallel, stage-wise. So first, it fetches all the data for any groups not already in the data cache in parallel using `rayon`, and then does the aligning in a second pass (also in parallel with rayon). For `coalesce_single`-eligible steps, we just do the whole coalesce-single job in the first phase, since there's no aligning to be done.

Benchmarks are widely variable now by what machine you run on, since core count now makes a difference given the automatic parallelism, but I'm seeing ~40% improvements over the previous version, which itself was ~60-70% better than upstream tree_coalesce depending on the benchmark.

I have a couple of minor passrate changes to dig into before merge, but I think this code, itself, is in good shape, except maybe for tweaking a couple of the constants used to decide where to limit processing.


rust-src/src/gridstore/coalesce.rs
```diff
@@ -1,12 +1,15 @@
 use std::borrow::Borrow;
-use std::cmp::Reverse;
+use std::cmp::{Ordering, Reverse};
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
+use std::sync::Arc;
```
aarthykc, line 6: @apendleton Not a blocker, curious about why you went with why Arc over RC?

rust-src/src/gridstore/coalesce.rs
```diff
@@ -332,121 +335,331 @@ fn coalesce_multi<T: Borrow<GridStore> + Clone>(
...
+    key_id: u32,
+    subquery: PhrasematchSubquery<T>,
+    key: MatchKey,
+    match_opts: MatchOpts,
+    is_single: bool,
+}
+
+// this is the thing that comes out of the first phase of two-phase coalesce
+// for single coalesce, we just do everything in phase 1, whereas for multi-coalesce,
+// we only do the first part, depending what kind of node we're on, we'll return different things
```
aarthykc, line 389: What do you mean by return different things here?

rust-src/src/gridstore/coalesce.rs
```diff
@@ -332,121 +335,331 @@ fn coalesce_multi<T: Borrow<GridStore> + Clone>(
...
+}
+
+fn penalize_multi_context(context: &mut CoalesceContext) {
+    // penalize single-entry stacks and ascending stacks for... some reason?
+    if context.entries.len() == 1 || context.entries[0].mask > context.entries[1].mask {
+        context.relev -= 0.01
+    }
+}
+
+pub const COALESCE_CHUNK_SIZE: usize = 8;
```
aarthykc, line 402: Is there a reason why we're doing 8 at a time?

rust-src/src/gridstore/coalesce.rs
```diff
@@ -332,121 +335,331 @@ fn coalesce_multi<T: Borrow<GridStore> + Clone>(
...
-
-                    contexts.push(context.clone());
-
-                    let state_vec = prev_state
-                        .entry((grid.grid_entry.x, grid.grid_entry.y))
-                        .or_insert_with(|| vec![]);
-                    state_vec.push(context);
+                // if this is a single item with no parents or children, we can do a more-efficient
+                // coalesce operation in the first phase, rather, than a two-phase coalesce
+                let is_single = step.prev_state.is_none() && step.node.children.len() == 0;
```
aarthykc, line 440: @apendleton For the data that we're fetching and is then available in the data_cache does that also include data that was pulled out for coalesce_single?

rust-src/src/gridstore/common.rs
```diff
@@ -385,6 +420,57 @@ pub struct PhrasematchSubquery<T: Borrow<GridStore> + Clone> {
     pub match_keys: Vec<MatchKeyWithId>,
 }
 
+pub struct ConstrainedPriorityQueue<T: Ord> {
```
aarthykc, line 423: These seem fine to me from reading examples in other places, but tagging in @CyanRook for another set of eyes!

rust-src/src/gridstore/coalesce.rs
apendleton: @aarthykc at the moment, in the first phase we *either* retrieve data for coalesce_multi (if it's a node with children and/or isn't at the base of the tree), or we just process the data in its entirety for coalesce_single (if it's at the base of the tree and childless). coalesce_single and coalesce_multi send different arguments to streaming_get_matching (specifically, coalesce_multi needs a ton more data), so if we're seeing it for the first time on coalesce_single and we only get a coalesce_single's worth of data, we'd need to do a subsequent request anyway to get the data needed for multi. At one point we talked a bit about this, and about flagging phrasematches so that we could know up-front when we encounter a phrasematch in a single context, like, "hey, at some point later, this same data is also going to be used for coalesce_multi," so that way we could go ahead and do the full multi fetch once instead of potentially doing two fetches for the same data, once for single and once for multi. As stackable doesn't provide that yet, though, that'll be left to a future optimization, and for now we'll live with the hit of potentially having to fetch twice.

rust-src/src/gridstore/coalesce.rs
apendleton: Just trial and error with benchmarks. Too many and we end up processing things we otherwise wouldn't have processed (like, maybe we'd already have enough results by the time we got there if we did them one at a time), so we end up wasting. But too few, and it's not parallel enough to get a performance improvement. 8 seemed to be the sweet spot, at least on a nemobox.

rust-src/src/gridstore/coalesce.rs
apendleton: @aarthykc `Arc` is the thread-safe version of `Rc`, so it's necessary if you want to share the same object across threads. The tradeoff is that it's slightly slower, which is why `Rc` also exists, so that you can use the fast version if you don't need thread safety.

rust-src/src/gridstore/coalesce.rs
apendleton: @aarthykc streaming_get_matching returns MatchEntry objects, and in coalesce we want to turn these eventually into CoalesceContexts. For coalesce_single, this is pretty much just a copy (we make a bunch of CoalesceContext objects that each have one entry apiece), and there's nothing else to be done -- we don't have to align things or calculate things or anything.

So here, if it's coalesce_single, we just go ahead and do the whole thing, all the way to creating a CoalesceContext, in the first phase, because there's no benefit to waiting and doing it later. For the multi ones, we'll need to align them and produce contexts out of data from multiple retrievals. So: the first phase produces *either* entire CoalesceContexts, for single-element stacks, or just the MatchEntry data, if it's going to go into multi-element stacks. That's why the result of this operation is an enum: one kind of output for single, the other kind for multi.

PR #84, opened 2020-04-21
Unified binned stackable

apendleton:
A couple of things in here:
- [x] merge StackableNode and StackableNodeCopy back together again
- [x] make `binned_stackable` be the recursive function and `stackable` be the wrapper function that does the binning, and sets all the tracking variables for recursion (the empty set, the zoom=0, etc.), so that now all the places that use stackable can just call `stackable(phrasematches)` and we don't have to change the call sites of like 150 coalesce fixtures every time we change the type signature of the recursive function
- [x] search-and-replace all the tests to just do the basic call
- [x] make the benchmark behavior match the upstream behavior (same sample size)

### DIFF
--- a/test_utils/src/lib.rs
+++ b/test_utils/src/lib.rs
@@ -11,7 +11,7 @@ use rusoto_core::Region;
 use rusoto_s3::{GetObjectRequest, S3Client, S3};
 use serde::{Deserialize, Serialize};
 
-use std::collections::{BTreeMap, HashMap, HashSet};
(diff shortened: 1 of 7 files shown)


test_utils/src/lib.rs
```diff
@@ -321,16 +321,3 @@ pub fn prepare_stackable_phrasematches(
         .collect();
     out
 }
-
-pub fn binned_phrasematches(
```
apendleton, line 325: This is moved inside stackable now

PR #85, opened 2020-04-22
[wip] Constrained stackable

apendleton:
This branch alters stackable to add limits to the size of the tree that can be generated. Salient specifics:

## ArenaManager

All of the storage for nodes has now been moved to a [generational arena](https://docs.rs/generational-arena/0.2.7/generational_arena/). This makes it much easier to have both a reference to a node inside the tree itself, and a reference to a node in a tracking/priority structure elsewhere. More specifically, there's now a wrapper type `ArenaManager` that contains an arena and also that bookkeeping infrastructure.

The new process for adding a child to an existing node, rather than adding the StackableNode to the `children` vector directly, is to add it to the arena, which will return an ID, and add *that* to the child node. Additionally, because the arena manager is keeping track of the limits on the size of the arena, adding a new node might fail. In that case, it is skipped rather than being added to the child list.

The way the bookkeeping works is like this: we limit on the maximum number of *leaves* (not nodes), so we enter into a constrained/full state (where we're potentially deleting things or limiting what we insert) when the maximum number of leaves is exceeded. Each node that gets added to the arena gets added to the tracking structure, but only increments the leaf count if it's a leaf. All nodes are ranked by the maximum possible relevance that can be attained on their subtree, or in other words, the highest-relevance leaf descended from that node (*not*, to be extra clear, the weight of that node's phrasematch, or the accumulated relevance so far up to that point).

The tracking structure is a HashMap pretending to be a kind-of-sort-of priority queue. Essentially, every max_relevance currently associated with a living node has an entry in the HashMap, and it maps to a tuple of a count of leaves with that (max_)relevance (for leaves, relevance and max_relevance are the same), and a list of the indexes of all of the nodes, leaf or otherwise, that have that relevance. We also keep track of the lowest relevance of any entry in the table.

We have a soft-max limit on the total number of leaves in the structure, so if we're under that limit when we add a node, all we do is add the node's index to the node list for its relevance, increment the leaf count if it's a leaf, and maybe reduce the min_relevance if necessary. If we're at or above the limit, things get more interesting:
* if the node we're trying to add is below the limit, we fail to add it
* if the node is exactly at the limit, we just do a regular add even though we're at the limit -- if the new thing we're trying to add is tied with the worst thing we have, we don't have any way to decide which to keep and which to toss out, so we just keep them both and let ourselves exceed the limit
* if the node is above the limit, we do a regular add, and then consider the possibility that even without the worst bin, we might now be at our limit. If that's the case, we delete the entire worst bin and all of its nodes in their entirety (so no having to choose between them) and then loop over the keys to pick a new worst one.

## max_possible_relev

The other chunk of changes is that we now calculate some more stuff on the bins of phrasematches introduced in #82: rather than just being vectors, they're now vectors plus some metadata. There are two metadata items: the highest weight of any phrasematch of this type, and the sum of the highest per-type weights for each type after this one. In other words: for the `place` type, I'll know the highest-weight `place` phrasematch, and the sum of the highest-weight `postcode` phrasematch plus the highest-weight `address` phrasematch. What that second sum makes it possible to do is to interrogate: okay, if I'm working on a stack, and I've accumulated a country, a region, and a place so far, and I know the total of their actual weights, what's the best possible weight that I could ever produce by stacking more stuff onto it, assuming I hit a homerun and got the best possible postcode and address? Combined with the limiting from the first work chunk, we can say as we're working: "okay, we know the lowest relevance the arena is willing to accept right now; if there's no possible way even with the best possible stacking that recursing further from this point will ever produce nodes with a high enough relevance to make it into the arena, we can just not bother to recurse down this path at all."

The first chunk got us memory savings, but not really speed savings, because we still computed everything, but just didn't keep some of it. This allows us to avoid computing things in the first place that we were guaranteed to throw away (it's not perfect -- the max possible relevance is overly optimistic most of the time, so we still *may* end up computing and then throwing away some stuff, but hopefully less of it).

## Other odds and ends

* `max_relevance` in treeify_stacks doesn't do what we expected it to, apparently -- it isn't the highest reachable relevance from a given position, which was leading to weird ranking decisions. That has now been corrected in this branch
* one odd side effect of culling things after they've already been added to the tree is the possibility that something can get added and then deleted later without its parent also being deleted, meaning nodes can now have entries in their `children` list pointing to things that no longer exist. I considered doing fancier bookkeeping so that nodes know who their parents are and, upon deletion, can remove themselves from their parents' child list, but ugh that was such a hassle. It ended up being easier to just tweak coalesce to skip over any nodes it encounters that are no longer present.

## Still to do

- [ ] tests
- [ ] maybe some more optimizations available... I'm curious about using SmallVec or TinyVec for the children list now that it's just a list of indexes, and possibly also for the bins in the arena tracking map. But going to profile and see where that gets me

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,6 +183,7 @@ dependencies = [
  "byteorder 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "criterion 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "generational-arena 0.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "integer-encoding 1.0.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "itertools 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lz4 1.23.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -538,6 +539,14 @@ version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "byteorder 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "generational-arena"
+version = "0.2.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


CyanRook:
If we need to come back and look for minor optimiziations, we should examine the arena.  @apendleton mentioned hashing being a large portion of the profiling time, but that could be with the hashset used outside the arena. It's unclear at this time.

apendleton:
This still needs better tests, but I'm going to land it for now in the interests of branches not diverging, and circle back on tests in a follow-up PR.

PR #86, opened 2020-04-23
Use fixedbitset instead of hashset

aarthykc:
## Context
Through profiling we realised that we spend a lot of time hashing - it could be from using hashsets for bmasks or from using them in the priority queues used to manage stackable. This PR switches hashsets for fixedbitset in attempts to improve performance. Since, it is not deserializ-able via the auto derive, we've chosen instead to deserialise it to a vec<u32> and then create the fixedbitmask. 

cc @apendleton @CyanRook

### DIFF
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -7,9 +7,9 @@ use neon::prelude::*;
 use neon::{class_definition, declare_types, impl_managed, register_module};
 use neon_serde::errors::Result as LibResult;
 use serde::Deserialize;
-use std::collections::HashSet;
 use owning_ref::OwningHandle;
 use failure::Error;
+use fixedbitset::FixedBitSet;
 
 use std::sync::Arc;
 
@@ -495,7 +495,8 @@ where
 
         let idx = js_phrasematch.get(cx, "idx")?;
 
-        let non_overlapping_indexes = js_phrasematch.get(cx, "non_overlapping_indexes")?;
+        let js_non_overlapping_indexes = js_phrasematch.get(cx, "non_overlapping_indexes")?;
+        let non_overlapping_indexes: Vec<u32> = neon_serde::from_value(cx, js_non_overlapping_indexes)?;
 
         let subq = PhrasematchSubquery {
             store: gridstore,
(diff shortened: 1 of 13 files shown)


native/src/lib.rs
```diff
@@ -506,13 +507,25 @@ where
...
+    for bit in bitset {
+        if bit == 1 {
+            fbs.insert(1);
+        } else {
+            fbs.insert(0);
+        }
+    }
+    return fbs;
+}
+
```
apendleton, line 528: Yeah, per chat, this would need to preserve the actual values, not just create a bitset with a 0 and a 1 in it, but as per above, I think you can just do the conversion inline anyhow rather than having a special function.

rust-src/src/gridstore/common.rs
```diff
@@ -414,7 +414,8 @@ pub struct MatchKeyWithId {
 pub struct PhrasematchSubquery<T: Borrow<GridStore> + Clone> {
     pub store: T,
     pub idx: u16,
-    pub non_overlapping_indexes: HashSet<u16>, // the field formerly known as bmask
+    #[serde(skip_serializing)]
```
apendleton, line 417: It sounded before like you had figured out the Serialize part (using `serialize_with` and `collect_seq`). Even if we can't do deserializing automatically because of the visitor nonsense, it would be good to do the serializing part, as if we ever want to produce bench data in the future, we'll need to be able to print out the contents of this field.

test_utils/src/lib.rs
```diff
@@ -30,7 +31,7 @@ pub fn round(value: f64, digits: i32) -> f64 {
 pub fn langarray_to_langfield(array: &[u32]) -> u128 {
     let mut out = 0u128;
     for lang in array {
-        out = out | (1 << *lang as usize);
+        out = out | (1 << *lang) as u128;
```
apendleton, line 34: Why did this change?

test_utils/src/lib.rs
```diff
@@ -11,6 +11,7 @@ use rusoto_core::Region;
 use rusoto_s3::{GetObjectRequest, S3Client, S3};
 use serde::{Deserialize, Serialize};
 
+use fixedbitset::FixedBitSet;
```
apendleton, line 14: super nitpick, but the imports are currently grouped with the internal stuff together, the third-party stuff together, and the standard library stuff together, so let's bump this up a line?

test_utils/src/lib.rs
```diff
@@ -249,13 +250,18 @@ pub fn prepare_phrasematches(
                                 .unwrap();
                                 Arc::new(gs)
                             });
+                        let mut fbs: FixedBitSet = FixedBitSet::with_capacity(128);
+
+                        for x in placeholder.non_overlapping_indexes.clone() {
+                            fbs.insert(x as usize);
+                        }
```
apendleton, line 257: as before, I think you could probably achieve this by collecting, rather than with the explicit loop here, but either way

native/src/lib.rs
```diff
@@ -506,13 +507,25 @@ where
             }],
             mask: neon_serde::from_value(cx, mask)?,
             idx: neon_serde::from_value(cx, idx)?,
-            non_overlapping_indexes: neon_serde::from_value(cx, non_overlapping_indexes)?,
+            non_overlapping_indexes: vector_to_fixedbitset(non_overlapping_indexes),
```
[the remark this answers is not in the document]
apendleton: I'm now realizing that maybe it's this way to be able to do the with_capacity thing?

PR #89, opened 2020-04-30
Allow phrasematches to be constrained to only nearby results

apendleton:
This branch introduces a new attribute on phrasematches, `nearby_only`, which allows Carmen to communicate that a phrasematch is likely to be expensive to query globally, and to only allow results that are close to the proximity point for that match.

On the Rust side, we currently have a hard-coded search radius for these phrasematches (25 miles), and calculate the number of tiles at that phrasematch's index's zoom level necessary to cover that radius, then calculate a bounding box by padding out the proximity point by that number of tiles in all directions (hard-stopping on the min side at 0 -- this is not antemeridian-crossing, but oh well). If the query already has a bbox, we intersect our newly generated bbox with the supplied one (so, only use the area covered by both).

Things that still need to happen:
- [x] add tests
- [x] fix the benchmarks -- because of this new field, they're currently all broken because none of our bench data includes it; I think I can tweak the benchmarks for which proximity doesn't matter to just assume the property is false (so that's all the stackable ones, and the non-prox coalesce ones), but will need to regenerate the data for the with-prox coalesce tests

### DIFF
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -424,7 +424,8 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
         // as long as there's still work to do, we'll execute it a chunk at a time, peeling off
         // the next few best nodes, and executing on them in parallel
         let mut step_chunk = Vec::with_capacity(COALESCE_CHUNK_SIZE);
-        let mut keys = HashMap::new();
+        let mut keys = Vec::new();
+        let mut unique_keys = HashSet::new();
         for _i in 0..COALESCE_CHUNK_SIZE {
             if let Some(step) = steps.pop_max() {
                 // if we've already gotten as many items as we're going to return, only keep processing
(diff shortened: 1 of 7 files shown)


apendleton:
The benchmarks run now, but they don't take advantage of this new functionality. We should regenerate the data that powers them to include this new property, but I think we might as well wait since we're going to be adding more properties anyway, most likely.

apendleton:
Oh, there's also a change slipped in here for slightly better deduping before the lookup phase in coalesce -- improves performance in a few edge cases.

rust-src/src/gridstore/coalesce.rs
```diff
@@ -425,6 +425,7 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
         // the next few best nodes, and executing on them in parallel
         let mut step_chunk = Vec::with_capacity(COALESCE_CHUNK_SIZE);
         let mut keys = Vec::new();
+        let mut unique_keys = HashSet::new();
```
aarthykc, line 428: This is to be able to dedup better, correct?

rust-src/src/gridstore/common.rs
```diff
@@ -111,6 +111,9 @@ impl Default for MatchOpts {
     }
 }
 
+pub const EARTH_CIRC_IN_MILES: f64 = 24901.0;
+pub const NEARBY_RADIUS: f64 = 25.0;
```
aarthykc, line 115: Wasn't this 10miles previously?

aarthykc:
@apendleton looks good to me pending the conflict. Also, wanted to ask what happens if the query has a proximity point and a bounding box?

rust-src/src/gridstore/common.rs
apendleton: Correct. 25 seems plenty fast and means we don't have any passrate regressions as compared to bounding box at all, in my tests.

rust-src/src/gridstore/coalesce.rs
apendleton: Yup, ensures that we don't fetch the same item more than once in the same fetch round.

apendleton:
@aarthykc in that case we calculate a bbox within 25 miles of the prox point, and then find the intersection of the calculated bbox with the bbox supplied by the user (in other words, any results will have to be in both): https://github.com/mapbox/carmen-core/pull/89/files#diff-2886181f7326a4ef3543698158b821cfR199-R206

PR #90, opened 2020-05-01
Single character

aarthykc:
## Context

After running a few benchmarks with slow queries, we realised that queries which are multi word and ending in a single character as slower in comparison with carmen-core master. This PR adds the ability to flag if fetching data for an index for such queries is particularly slow and using that information to reduce the number of fetches. 

## Next steps 
- [x] Needs a lot of testing to confirm whether it's actually going to be helpful in single character multi word cases 
- [x] Review
- [ ] Merge with `treeify-stacks`

cc @apendleton @CyanRook

### DIFF
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -619,7 +619,8 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
                                         new_context.mask = new_context.mask | subquery.mask;
                                         new_context.relev += entry.grid_entry.relev;
                                         if new_context.relev > step.relev_so_far {
-                                            step.relev_so_far = new_context.relev;
+                                            step.relev_so_far = new_context.relev
+                                                + step.node.phrasematch.expect("pm").weight;
                                         }
 
                                         let mut out_context = new_context.clone();
@@ -654,7 +655,8 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
                                 };
 
                                 if context.relev > step.relev_so_far {
-                                    step.relev_so_far = context.relev;
+                                    step.relev_so_far =
+                                        context.relev + step.node.phrasematch.expect("pm").weight;
                                 }
 
                                 let mut out_context = context.clone();
@@ -673,16 +675,14 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
                     let mut next_steps = Vec::with_capacity(step.node.children.len());
                     if state.len() > 0 {
                         let state = Arc::new(state);
-                        let mut relev_so_far = 0.0;
                         for child_idx in step.node.children.iter() {
                             if let Some(child) = stack_tree.arena.get(*child_idx) {
-                                relev_so_far += child.phrasematch.expect("pm").weight;
                                 next_steps.push(CoalesceStep::new(
                                     &child,
                                     Some(state.clone()),
                                     subquery.store.borrow().zoom,
                                     match_opts,
-                                    relev_so_far,
+                                    step.relev_so_far,
                                 ));
                             }
                         }
@@ -721,9 +721,8 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
                     step.relev_so_far + step.node.phrasematch.expect("require phrasematch").weight;
                 if step.node.is_leaf()
                     && step.node.zoom == 14
-                    && step_relev
-                        <= (0.5 * contexts.peek_max().expect("contexts can't be empty").relev)
-                    && counter == 1
+                    && step.relev_so_far.min(1.0)
+                        <= contexts.peek_max().expect("contexts can't be empty").relev
                     && count_bits(step.node.phrasematch.expect("expect pm").mask) == 1
                 {
                     continue;
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -675,7 +675,7 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
                                     Some(state.clone()),
                                     subquery.store.borrow().zoom,
                                     match_opts,
-                                    relev_so_far
+                                    relev_so_far,
                                 ));
                             }
                         }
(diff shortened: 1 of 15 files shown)


rust-src/src/gridstore/coalesce.rs
```diff
@@ -784,6 +836,17 @@ pub fn collapse_phrasematches<T: Borrow<GridStore> + Clone + Debug>(
     phrasematch_results
 }
 
+pub fn count_bits(mut mask: u32) -> u32 {
```
apendleton, line 839: FYI there's a built-in function to do this: https://doc.rust-lang.org/std/primitive.u64.html#method.count_ones

rust-src/src/gridstore/store.rs
```diff
@@ -27,6 +27,7 @@ pub struct GridStore {
     pub zoom: u16,
     pub type_id: u16,
     pub coalesce_radius: f64,
+    pub is_slow: Option<bool>,
```
apendleton, line 30: Hm, why is this an option<bool> and not just a bool?

aarthykc: Changed!

rust-src/src/gridstore/coalesce.rs
aarthykc: Ha, thanks for pointing out -- changed.

rust-src/src/gridstore/coalesce.rs
```diff
@@ -548,7 +550,7 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
         let chunk_results: Vec<Result<(Vec<CoalesceContext>, Vec<CoalesceStep<'_, T>>), Error>> =
             step_chunk
                 .into_par_iter()
-                .map(|step| {
+                .map(|mut step| {
```
apendleton, line 553: I don't think we need to mutate the step we're on now -- that's the accumulated relevance from the last round. We just need to keep track of the relevance so far of this new round for the next step, so I think this could stay `|step|` and then start over again with a fresh count below it (`let mut relev_so_far = 0.0` or whatever)

rust-src/src/gridstore/coalesce.rs
```diff
@@ -591,6 +593,9 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
 
                                         new_context.mask = new_context.mask | subquery.mask;
                                         new_context.relev += entry.grid_entry.relev;
+                                        if new_context.relev > step.relev_so_far {
```
apendleton, line 596: then this would be `if new_context.relev > relev_so_far`, correspondingly

apendleton: (and same below)

rust-src/src/gridstore/coalesce.rs
```diff
@@ -697,7 +708,42 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
...
+                for key in &step.node.phrasematch.expect("phrasematch required").match_keys {
+                    let start = match key.key.match_phrase {
+                        MatchPhrase::Exact(phrase_id) => phrase_id,
+                        MatchPhrase::Range { start, .. } => start,
+                    };
+
+                    let end = match key.key.match_phrase {
+                        MatchPhrase::Exact(phrase_id) => phrase_id,
+                        MatchPhrase::Range { start: _, end } => end,
+                    };
```
apendleton, line 723: Can we make these a single match? e.g., `let (start, end) = ...` and then in the match arms, output both as a tuple

rust-src/src/gridstore/coalesce.rs
```diff
@@ -697,7 +708,42 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
...
+
+                    let range = end - start;
+                    if range > 1 {
+                        counter = 1;
```
apendleton, line 727: we're not counting anything, right? let's just make this a boolean -- is_range or something maybe?

rust-src/src/gridstore/coalesce.rs
```diff
@@ -697,7 +708,42 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
             for context in phrasematch_contexts {
                 contexts.push(context);
             }
+
             for step in next_steps {
+                let mut counter = 0;
+                for key in &step.node.phrasematch.expect("phrasematch required").match_keys {
```
apendleton, line 714: over the course of this loop when end up doing this dance of grabbing the phrasematch and doing the expect thing like three times... probably worth doing a `let phrasematch =...` at the top of the loop and making all the subsequent uses (733, 743, etc.) shorter

rust-src/src/gridstore/coalesce.rs
```diff
@@ -697,7 +708,42 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
...
+                if step.node.is_leaf()
+                    && step
+                        .node
+                        .phrasematch
+                        .expect("phrasematch required")
+                        .store
+                        .borrow()
+                        .might_be_slow
+                    && step.relev_so_far
+                        <= 0.75 * contexts.peek_max().expect("contexts can't be empty").relev
```
apendleton, line 741: I'm not 100% sure there will necessary always be things in contexts, and if there aren't, this will panic. I think you could instead do something like [this](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or) (haven't actually tried it), but I think it would look something like `contexts.peek_max().map_or(0.0, |m| m.relev)`, which would give you either the relev if there's a relev, or 0 (and since no new context will have a relev <= 0, this check would let everything through, which I think would be what you would want).

rust-src/src/gridstore/store.rs
```diff
@@ -256,6 +258,10 @@ impl GridStore {
         GridStore::new_with_options(path, 6, 0, 0.0, [0, 0, 63, 63])
     }
 
+    pub fn mark_slow_indexes(zoom: u16) -> bool {
```
apendleton, line 261: Hm, why not make this a method on an actual gridstore, rather than a static method? Like, have it just take `&self` instead of a zoom, and look at self.zoom

rust-src/src/gridstore/store.rs
```diff
@@ -286,7 +292,17 @@ impl GridStore {
...
+        let might_be_slow = GridStore::mark_slow_indexes(zoom);
+        Ok(GridStore {
+            db,
+            path,
+            bin_boundaries,
+            zoom,
+            type_id,
+            coalesce_radius,
+            bbox,
+            might_be_slow,
```
apendleton, line 304: and then if we did the method, we wouldn't need to store the value, we could just call the method

PR #91, opened 2020-05-06
Bbox aware coalesce

apendleton:
This PR now allows for indexes to, when they construct their gridstore, specify the bounding box (in tile coordinates at the given index's zoom level) of the data in the index.

In tree coalesce, as we coalesce each layer we now keep track of the bounding box of the tiles we've successfully aligned, and before deciding to explore and fetch a child's data, we confirm that the bounding box of the index corresponding to the child's phrasematch overlaps with the working bounding box that bounds the current layer's data.

For a concrete example: supposing we have a global region layer containing an "Oregon" feature, and geographically constrained indexes we're stacking on top of it, previously after fetching the region data and deciding whether to stack other indexes on top of it, those indexes only had to overlap with the region layer generally (so, because the region layer is global, this would include all indexes). Now, it only attempts fetching and stacking if the child layer overlaps with the bounds of all of the features called "oregon," which potentially excludes indexes for features outside the US.

### DIFF
--- a/native/src/lib.rs
+++ b/native/src/lib.rs
@@ -81,6 +81,7 @@ struct GridStoreOpts {
     pub zoom: u16,
     pub type_id: u16,
     pub coalesce_radius: f64,
+    pub bbox: [u16; 4],
 }
 
 declare_types! {
@@ -286,7 +287,8 @@ declare_types! {
                         filename,
                         opts.zoom,
                         opts.type_id,
-                        opts.coalesce_radius
+                        opts.coalesce_radius,
+                        opts.bbox,
                     )
                 },
                 None => GridStore::new(filename)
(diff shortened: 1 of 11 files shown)


aarthykc:
@apendleton should we add a test in coalesce that demonstrates that we skip stacking if the bboxes don't intersect? Also not opposed to adding it come in a follow up PR -- just want to make sure we don't break this behaviour.

apendleton:
@aarthykc yeah, everything we've been adding is under-tested for sure. It's tricky to figure out how to test this one in particular, because the externally observable results don't change: if there wasn't any data in a particular area, we wouldn't return any results even without the bounding box, we would just have wasted to time first. In that sense, the current tests still passing is probably what confirms that things are working right. That said, none of the current tests actually specify bounding boxes that bound their actual data, so in a follow-up PR I'll plan to add some tests that do that.

PR #94, opened 2020-07-09
Smarter bbox coalesece + quotas for single-letter subqueries

apendleton:
This PR adds a couple of new things, one general-purpose and a couple specific to single-letter-word autocomplete

## General purpose: smarter bbox filtering

This branch makes a couple of improvements on the bbox-aware coalesce mechanism introduced in #91. First, rather than each index having a single bounding box, it can now have multiple, which are considered to be unioned together for purposes of filtering. Concretely, this makes representation of antemerdian-crossing bounding boxes possible in a way that doesn't require them to span the whole world, by splitting said bounding boxes in two at the antemeridian; this is what we now do on the Carmen branch corresponding to this one, though it should be noted that in theory, the new schema would allow for further refinement, e.g., by drawing distinct boxes around each polygon in a multipolygon, which could create further efficiency improvements beyond what we're seeing now without further changes in this repo.

Next, the running bounding box comparisons during coalesce have been reworked. Before, at each coalesce layer, we calculated the bounding box of all the tiles we'd seen corresponding to the current subquery in the current index, and when deciding whether to recurse and stack an additional layer atop the current one, we would check to see if the index corresponding to the next candidate subquery had a bounding box that overlapped with our calculated bounding box before deciding whether to recurse. This was better than nothing (what we were doing before), but some queries would generate bounding boxes that were not that useful for purposes of excluding child candidates. For example, a city subquery with matches in both Canada and the UK would produce a large bounding box that would cover a large portion of the north Atlantic. When considering whether to stack, say, Ireland addresses on top of this subquery, we'd pass the bounding box test even though we hadn't actually found any individual city tiles in Ireland.

The new mechanism replaces the xy-keyed dictionary we had been using to store state at each subquery layer (a strategy we inherited from carmen-cache) with a proper spatial index: a Flatbush from https://github.com/apendleton/static-bushes, which itself is a port from @mourner 's https://github.com/mourner/flatbush library. This allows us to do the same dictionary lookups we were doing before, but to replace the bounding box overlap calculation with a better mechanism: we can look at the bounding box (or, with the bounding box decomposition changes above, multiple bounding boxes) of the candidate index we're proposing to stack, and check if it overlaps not just with the bounding box of the tiles we've seen so far, but instead check if it contains any of the actual tiles. In the Canada/UK/Ireland example above, given an Irish address index whose bounds matched the country of Ireland, and given that those bounds wouldn't contain any actual tiles we had found in Canada or the UK, we'd be able to exclude Irish addresses as a potential child match.

## Specific to slow queries
- [x] **move where the limit introduced in #90 occurs:** #90 adds a mechanism for not considering some subqueries if we think they'll be slow based on heuristics about the index, and if it's not possible for them to generate results with a relevance that's at least close to the relevance of the best thing we've seen so far. In that PR, however, we did that check when we queued the index for possible later processing. By moving the check to right before we actually process the subquery, in many cases the "best thing we've seen so far" has gotten a lot better in the meantime as we've processed more stuff, so the filter ends up being able to exclude more things.
- [x] **add explicit tracking of how long each phrase is to allow special treatment of one-character subqueries**
- [x] **add an explicit quota for these subqueries:** we now count how may of these really-slow queries (one-character) we're willing to process, and relevance notwithstanding, at a certain point we just stop doing them. Currently this quota is set to 8 per query. Note that this is a limit on fetches, not instances of stacking, so if we've already fetched a given subquery when we next encounter it, we can still process it from cache. We just won't fetch new ones.
- [x] **improve ranking of subqueries for processing:** prior to this PR, the mechanism for deciding the order of stacking operations was a simple priority queue keyed just on the maximum possible relevance a given subtree could generate. This generated lots of ties, but in practice that didn't matter because we'd process all of them anyway. Now that we're introducing quotas, having a smarter ranking becomes more important so we can decide which things make the cut. This PR adds information about the maximum score of any feature in a given index to the gridstore metadata, and uses that as an additional tie-breaker to decide what to stack next.

## Misc
- [x] **add `Default` implementations for some types:** each time we add a new field to a type, we've had to go in and update every instantiation of that type in tests, and there are sometimes hundreds. This PR adds `Default` implementations for at least some of them, and uses those implementations in tests so that hopefully in the future when we add more fields to these types we won't have to update as many things. We should do this for more types, but that should probably be a follow-up PR.

## Next actions
- [ ] the benches are still broken and need regenerated data

### DIFF
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -933,7 +933,7 @@ mod test {
         builder.insert(&key, entries).expect("Unable to insert record");
         builder.finish().unwrap();
         let store1 =
-            GridStore::new_with_options(directory.path(), 14, 1, 200., global_bbox_for_zoom(14))
+            GridStore::new_with_options(directory.path(), 14, 1, 200., global_bbox_for_zoom(14), 1.0)
                 .unwrap();
 
         let a1 = PhrasematchSubquery {
@@ -943,9 +943,9 @@ mod test {
             weight: 0.5,
             mask: 1,
             match_keys: vec![MatchKeyWithId {
-                nearby_only: false,
                 key: MatchKey { match_phrase: Range { start: 0, end: 1 }, lang_set: 0 },
                 id: 1,
+                ..MatchKeyWithId::default()
             }],
         };
 
@@ -956,9 +956,9 @@ mod test {
             weight: 0.5,
             mask: 1,
             match_keys: vec![MatchKeyWithId {
-                nearby_only: false,
                 key: MatchKey { match_phrase: Range { start: 0, end: 1 }, lang_set: 0 },
                 id: 2,
+                ..MatchKeyWithId::default()
             }],
         };
         let phrasematch_results = vec![a1, a2];
--- a/rust-src/src/gridstore/coalesce.rs
+++ b/rust-src/src/gridstore/coalesce.rs
@@ -6,12 +6,12 @@ use std::fmt::Debug;
 use std::sync::Arc;
 
 use failure::Error;
-use static_bushes::{FlatBush, FlatBushBuilder};
 use indexmap::map::{Entry as IndexMapEntry, IndexMap};
 use itertools::Itertools;
 use min_max_heap::MinMaxHeap;
 use ordered_float::OrderedFloat;
 use rayon::prelude::*;
+use static_bushes::{FlatBush, FlatBushBuilder};
 
 use crate::gridstore::common::*;
 use crate::gridstore::spatial::adjust_bbox_zoom;
@@ -382,10 +382,7 @@ impl<T: Borrow<GridStore> + Clone + Debug> CoalesceStep<'_, T> {
     #[inline(always)]
     fn cmp_key(&self) -> (OrderedFloat<f64>, OrderedFloat<f64>) {
         let subquery = self.node.phrasematch.expect("phrasematch required");
-        (
-            OrderedFloat(self.node.max_relev),
-            OrderedFloat(subquery.store.borrow().max_score)
-        )
+        (OrderedFloat(self.node.max_relev), OrderedFloat(subquery.store.borrow().max_score))
     }
 }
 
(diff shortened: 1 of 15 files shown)


rust-src/src/gridstore/coalesce.rs
```diff
@@ -562,13 +624,16 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
...
+                    let mut state_contexts: Vec<CoalesceContext> = Vec::new();
 
                     for key_group in subquery.match_keys.iter() {
-                        let grids = data_cache
-                            .get(&key_group.id)
-                            .expect("data must have been pre-collected");
+                        let grids = match data_cache.get(&key_group.id) {
+                            Some(data) => data,
+                            None => {
+                                // we must have skipped collecting this data
```
aarthykc, line 633: When would we skip collecting the data?

rust-src/src/gridstore/coalesce.rs
```diff
@@ -459,6 +490,37 @@ pub fn tree_coalesce<T: Borrow<GridStore> + Clone + Debug + Send + Sync>(
...
+                                // this is a potentially-slow leaf subquery in a high-zoom index
+                                // that isn't likely to make our best results better, so skip it
+                                continue;
+                            } else if key_group.phrase_length == 1
+                                && !unique_keys.contains(&(key_group.id, is_single))
+                            {
+                                // even though this isn't a leaf node or a high-zoom index, it's a single-letter query,
+                                // which could be *really* slow and is pretty low-information, so set a quota to constrain
+                                // the total number of these we can end up fetching
+                                if short_range_count < SHORT_RANGE_QUOTA {
```
aarthykc, line 516: Nice!

rust-src/src/gridstore/coalesce.rs
apendleton: because it was over quota when we did the data collection -- that decision happened in the previous phase, so if we went get to this phase, it didn't happen, it must be because in the previous phase we decided not to

rust-src/src/gridstore/spatial.rs
```diff
@@ -598,10 +598,6 @@ pub fn scoredist(mut zoom: u16, mut distance: f64, mut score: u8, radius: f64) -
     ((6. * E_POW[score as usize] / E_POW[7]) + 1.) / dist_ratio
 }
 
-pub fn bboxes_intersect(bbox1: [u16; 4], bbox2: [u16; 4]) -> bool {
```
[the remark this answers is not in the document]
apendleton: Yup, correct. The previous logic was "does the bounding box of the index we're considering stacking on top of this set of tiles overlap with the bounding box of the set of tiles we found in this index" and the new logic is "does the bounding box of the index we're considering stacking on top of this set of tiles contain any of the specific tiles from this set," so there's not a bounding-box-overlap operation anymore (and the compiler complains that that function isn't being used anymore if I leave it in place).

rust-src/src/gridstore/common.rs
```diff
@@ -445,6 +451,20 @@ pub struct MatchKeyWithId {
...
+    pub phrase_length: usize,
+}
+
+impl Default for MatchKeyWithId {
+    fn default() -> Self {
+        MatchKeyWithId {
+            key: MatchKey::default(),
+            nearby_only: false,
+            id: 0,
+            // default is 2 because 1 has special behaviors that we might not want to opt into
```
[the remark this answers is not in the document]
apendleton: The short-query quota potentially kicks in for length=1. For the typical test case where we want "regular" behavior, better to do length>1 to be on the safe side.

PR #98, opened 2021-02-17
Modernize carmen-core, plus merge in node-fuzzy-phrase

apendleton:
This branch supports getting Carmen running on Node 14, along with some other long-delayed cleanup. We're also going to deprecate the node-fuzzy-phrase repo (the Node wrapper for fuzzy-phrase), and wrap fuzzy-phrase (the pure-rust library) here instead. Eventually this will facilitate tying these more tightly together, and hiding more of the interop between the two from Node, but for now both the current GridStore and FuzzyPhraseSet will be exposed as-is, just from a single library instead of two.

Specifically:
- [x] upgrade carmen-core from Neon 0.2 to 0.7
- [x] move CI to Node 14 and Ubuntu 20.04 (note: newly built modules will now no longer work on old Node or old Ubuntu)
- [x] stitch in the history of node-fuzzy-phrase (and so, introduce a dependency on fuzzy-phrase)
- [x] port fuzzy-phrase bindings from Neon 0.1 to Neon 0.7 (this is effectively a rewrite -- Neon 0.1 to 0.2 was a significant change)
- [x] delete never-used carmen-cache submodule

Still to do:
- [x] get all downstream tests to pass


index.js
```diff
@@ -7,4 +7,10 @@ addon.GridStore.prototype.keys = function() {
     return out;
 }
 
+addon.ENDING_TYPE = {
+    nonPrefix: 0,
+    anyPrefix: 1,
+    wordBoundaryPrefix: 2,
+}
```
apendleton, line 14: This was formerly in node-fuzzy-phrase

native/Cargo.toml
```diff
@@ -9,17 +9,19 @@ edition = "2018"
...
 [build-dependencies]
-neon-build = "0.2.0"
+neon-build = "0.7.1"
 
 [dependencies]
-neon = "0.2.0"
-neon-serde = "0.1.1"
+neon = "0.7.1"
+# neon-serde branch that supports neon 0.7, until https://github.com/GabrielCastro/neon-serde/pull/67 lands
+neon-serde = { git = "https://github.com/lucasfernog/neon-serde.git", rev="a9d9164f45ff65f6f64037fe8cb3490bed357580" }
```
apendleton, line 20: Maintainership of neon-serde seems a little unclear, but this is still not merged. I'm thinking probably we just need to live with shipping a build from a fork for the time being.

native/src/fuzzy_phrase.rs
```diff
@@ -0,0 +1,308 @@
+use neon::declare_types;
```
apendleton, line 1: The diff on this file shows it all being new, but it's actually, firstly, a copy into this repo of https://github.com/mapbox/node-fuzzy-phrase/blob/master/native/src/lib.rs, and secondly, a patch to rewrite it for Neon 0.7. To get a sense of what's actually changed, see the commit that does the porting: https://github.com/mapbox/carmen-core/pull/98/commits/0c52f107e13e311d34b4f0b9b517ce6e95987c90#diff-6da75502e50ffa41c6f57ee798d1dc9150cdb84690d4a5af8057d0869ca13062 -- but it's a major rework, as the API from Neon 0.1 to 0.7 has changed very significantly.

native/Cargo.toml
aarthykc: Agreed, I feel like we might have done that before as well?

.gitmodules
```diff
@@ -1,6 +0,0 @@
-[submodule "carmen-cache-sys/carmen-cache"]
```
aarthykc, line 1: nice!

.travis.yml
```diff
@@ -72,7 +75,7 @@ matrix:
         - source "$HOME/.cargo/env"
         - yarn install --ignore-scripts
         - neon clean
-        - neon build $([ "${BUILDTYPE}" == 'debug' ] && echo "" || echo "--release")
+        - travis_wait 30 neon build $([ "${BUILDTYPE}" == 'debug' ] && echo "" || echo "--release")
```
aarthykc, line 78: Out of curiosity, what is this for?

tests/bindings/test.js
```diff
@@ -167,7 +167,7 @@ tape('Coalesce tests - invalid inputs', (t) => {
         zoom: 6,
```
aarthykc, line 167: [nit] Should we rename this file to indicate that it tests the gridstore node bindings?

aarthykc (approved):
Looks good to me, very excited about the node binding consolidation!

.travis.yml
apendleton: Ah right, forgot to note this. By default, Travis kills a build job if it doesn't produce any output for 10 minutes, and for some reason, OS X builds on Travis just take longer now than they used to, and were getting killed before they could complete. Building locally on OS X doesn't seem subjectively slower to me than it ever has, so maybe Travis has gotten stingier on the hardware being used for OS X build workers? Not clear. Either way: the `travis_wait` thing just keeps the task from getting killed for 30 minutes.

tests/bindings/test.js
apendleton: Good call, renamed.

PR #103, opened 2021-10-29
Support up to 256 indexes

michaelskahn:
## Summary of changes

This PR expands index support in carmen-core from 128 to a maximum of 256. There's two changes needed here to implement that feature:

- Use 8 instead of 7 bits for the index component of the `tmp_id`. This is safe, because the feature ID component of the `tmp_id` uses only the lower 24 bits.
- Change the size of the `non_overlapping_indexes` `FixedBitSet` from 128 to 256.

This update must be paired with a corresponding update to carmen (https://github.com/mapbox/carmen/pull/973) so that `tmpid` calculation is consistent. 

## Testing

All tests in carmen-core and carmen continue to pass with updates to the `tmpids` in fixtures. There is no detectable impact on result quality when <128 indexes are used.

Tests that previously failed with `128 < num_indexes < 256` now pass.

carmen-core benchmark results:
<details>

```
coalesce_global         time:   [2.3540 ms 2.3739 ms 2.3991 ms]
                        change: [-18.199% -10.456% +1.0263%] (p = 0.06 > 0.05)
                        No change in performance detected.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild

coalesce_prox           time:   [2.8152 ms 2.8479 ms 2.8888 ms]
                        change: [-2.5445% +1.4978% +5.4161%] (p = 0.48 > 0.05)
                        No change in performance detected.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild

coalesce_bounds         time:   [2.4048 ms 2.4233 ms 2.4467 ms]
                        change: [-2.6905% +0.7326% +4.4086%] (p = 0.70 > 0.05)
                        No change in performance detected.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild

coalesce_ac_global      time:   [5.2274 ms 5.5625 ms 5.8132 ms]
                        change: [+16.906% +63.107% +143.04%] (p = 0.01 < 0.05)
                        Performance has regressed.
Found 2 outliers among 20 measurements (10.00%)
  2 (10.00%) low mild

coalesce_ac_prox        time:   [9.7526 ms 9.9418 ms 10.196 ms]
                        change: [-7.1583% -1.0439% +5.6789%] (p = 0.75 > 0.05)
                        No change in performance detected.
Found 5 outliers among 20 measurements (25.00%)
  1 (5.00%) low severe
  2 (10.00%) low mild
  2 (10.00%) high mild

coalesce_ac_bounds      time:   [5.3890 ms 5.7412 ms 5.9960 ms]
                        change: [-21.076% -0.6498% +22.024%] (p = 0.95 > 0.05)
                        No change in performance detected.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) low mild

stackable_us_address    time:   [567.13 us 619.51 us 650.68 us]
                        change: [-43.083% -26.340% -8.7003%] (p = 0.01 < 0.05)
                        Performance has improved.

stackable_us_address_postcode
                        time:   [355.86 us 371.93 us 384.18 us]
                        change: [-24.912% -3.0887% +24.534%] (p = 0.81 > 0.05)
                        No change in performance detected.

stackable_us-address_postcode_place_region
                        time:   [1.3610 ms 1.4475 ms 1.4971 ms]
                        change: [-21.331% -1.4039% +22.760%] (p = 0.89 > 0.05)
                        No change in performance detected.
```
</details>

There's a performance regression for the `coalesce_ac_global` benchmark. We'll have to see if this has an impact on overall latency.

/cc @allisonzheng @mattciferri


tests/coalesce_test.rs
```diff
@@ -551,7 +551,7 @@ fn coalesce_single_test() {
                 phrasematch_id: 0,
                 matches_language: true,
                 idx: 1,
-                tmp_id: 33554433,
+                tmp_id: 16777218,
```
allisonzheng, line 554: I see that the `tmp_id`s are consistently `16777218` after the changes but how come the `tmp_id`s were different before?

michaelskahn: Good question. They aren't *all* the same, but they don't consistently reflect the differences that were there before. I'll take a closer look and see if I can figure out why.

michaelskahn: Some tests were testing `CoalesceContext` equality, which is less strict than `CoalesceEntry` equality, and didn't take the `tmp_id` into account. I've updated to make some of these tests stricter, and set `tmp_id` to 0 in the others to make it more clear that the value isn't relevant.

mattciferri (approved):

